PR #7, opened 2018-05-16
Avoid having to specify GLSLANG_REPO_ROOT when running CMAKE

mark-lunarg:
This change will download the glslangValidator binary automatically, avoiding external dependencies upon a separate glslang repo.  Build.md doc updated with these instructions, along with how to set a location manually if a different glslangValidator version is wanted (for some crazy reason).

Note: The only glslang component currently used in this repo is glslangValidator, used to compile the two dinky shaders in cube and cubepp.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ CMakeLists.txt.user
 CMakeFiles/
 cmake_install.cmake
 Makefile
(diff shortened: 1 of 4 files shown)


mark-lunarg:
Pushed to Master.

PR #9, opened 2018-05-17
Add Build Capability for MacOS

jeremyk-lunarg:
Updated CMake scripts to allow building on MacOS.
Updated BUILD.md to reflect the change in build workflow for MacOS.


jeremyk-lunarg:
If there is no further input by the end of the day, I will rebase this PR.

PR #24, opened 2018-05-29
cube: Use time-based animation

joey-lunarg:
Update the cube's angle based on the elapsed time and angle-per-second,
instead of a fixed angle per frame.

Change-Id: Iffe33098203c27045798e85277a04c5f28de7da9

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -26,6 +26,7 @@
  */
 
 #define _GNU_SOURCE
+#define _CRT_NO_TIME_T
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


karl-lunarg:
`cubepp` seems to run OK, but for me, `cube` rotates really slowly, like about less than 1 degree/sec.  (Ubuntu 16.04)

karl-lunarg:
`cube` looking better now.  I still need to review code.

jeremyk-lunarg:
@karl-lunarg @dustin-lunarg Does this still look good?  Can we merge this or does it still need changes?

jeremyk-lunarg:
Determined to add unwanted complexity.
Closing issue.

PR #27, opened 2018-06-01
Clean up validation errors from separate present queue

Tony-LunarG:
From the spec:

Whilst it is not invalid to provide destination or source access
masks for memory barriers used for release or acquire operations,
respectively, they have no practical effect. Access after a release
operation has undefined results, and so visibility for those accesses
has no practical effect. Similarly, write access before an acquire
operation will produce undefined results for future access,
so availability of those writes has no practical use. In an earlier
version of the specification, these were required to match on both
sides - but this was subsequently relaxed.
These masks should be set to 0.

Change-Id: I495dc86ad62c0651fbc6acbfb0dfbb8245a324be


karl-lunarg:
This is failing CI because the CI scripts need to be updated as a result of Vulkan-Loader needing to be built differently.  I'm trying to fix this in #30.

karl-lunarg:
I fixed the clang-format and commit format problems.  Unfortunately, Travis-CI doesn't seem to be getting triggered for new PRs, so I can't fix the CI problems now.  But this is now safe to merge.

PR #31, opened 2018-06-05
build: Adjust ci for updated Vulkan-Loader

karl-lunarg:
Adapt to Vulkan-Loader getting rid of Vulkan-Headers submodule.
Get and build Vulkan-Headers and fix Vulkan-Loader build to use it.


karl-lunarg:
Travis jobs are not being scheduled.

PR #34, opened 2018-06-05
repo: Reorg to remove Vulkan-Headers submodule

karl-lunarg:
- Remove Vulkan-Headers submodule
- Add CMake code to add VULKAN_HEADERS_INSTALL_DIR and
  VULKAN_LOADER_INSTALL_DIR to search when locating
  Vulkan headers and loader

(LOADER_REPO_ROOT still being honored for now)


karl-lunarg:
Yes, it actually needs to set both HEADERS and LOADER.  I'll update it.  Thanks.

PR #35, opened 2018-06-06
android: Fix CI for Android

karl-lunarg:
Call `update_external_sources_android.sh` in the Travis job.
Add to this script the ability to fetch Vulkan-Headers repo.
Update the location of the Vulkan headers to match the fetched location.


karl-lunarg:
@cnorthrop I added sync switches to the Linux version of the update script so that it is easy to select which repos are fetched.  I then left only the Vulkan-Headers fetch turned on to reduce the number of clones and network traffic.  I also updated both scripts to handle additional repos that are available.  But their fetches are disabled via the switches I added.

This change seems to build Cube successfully.

PR #38, opened 2018-06-12
Temporarily rename uninstall target

mark-lunarg:
Rename uninstall target until downstream repos have been modified to remove submodules.  

Target is renamed to uninstall-Vulkan-Tools, this commit can be reverted once the VulkanTools repository rework is complete.

This is required to get VulkanTools working with the new repos.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,14 +46,16 @@ if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND WIN32)
 endif()
 
 # uninstall target
-configure_file(
-    "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
-    "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
-    IMMEDIATE @ONLY)
+if (NOT TARGET uninstall)
+    configure_file(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -51,9 +51,9 @@ configure_file(
     "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
     IMMEDIATE @ONLY)
 
-add_custom_target(uninstall
+add_custom_target(uninstall-Vulkan-Tools
```
karl-lunarg: I think a more permanent and better solution is possible:

How about checking to see if the uninstall target is defined and if not, run all this code to define it.

That way, this repo can be included as a submodule by another repo that defines uninstall and only the top level repo defines uninstall, which is what you want.  Further, you don't have to invent odd "uninstall-repo-name" target names.

mark-lunarg: Ah, nice!  I'll give it a try. Do you recommend doing the same in other repos with uninstall targets, or just here?

karl-lunarg: We should do it to all 4 of the new repos and any others that we have that might be used as submodules.   I don't know the exact CMake syntax for checking this, but `if(NOT uninstall)` would be a good start.

mark-lunarg: if(NOT TARGET uninstall)

PR #50, opened 2018-07-09
cmake: Clean up macOS install target

mikes-lunarg:
- Stop hardcoding CMAKE_INSTALL_PREFIX to CMAKE_BINARY_DIR so install
  products are in a predictable location
- Use install/fixup_bundle to create stand-alone app bundles for cube,
  cubepp, vulkaninfo-bundle at CMAKE_INSTALL_PREFIX
- Set INSTALL_RPATH_USE_LINK_PATH for bundle targets so fixup_bundle can
  find libraries when run from the install tree
- Remove extraneous copy of libMoltenVK.dylib from CMAKE_INSTALL_PREFIX/demos
- Update BUILD.md


BUILD.md
```diff
@@ -738,40 +741,37 @@ To address this problem, run:
...
     vulkaninfo/vulkaninfo
 
-If you run this before you run "make install", vulkaninfo's RPATH is already
+If you run this from the build directory, vulkaninfo's RPATH is already
 set to point to the Vulkan loader in the build tree, so it has no trouble
 finding it. But the loader will not find the MoltenVK driver and you'll see a
 message about an incompatible driver. To remedy this:
 
-    VK_ICD_FILENAMES=<path-to>/MoltenVK/Package/Latest/MoltenVK/macOS/MoltenVK_icd.json demos/vulkaninfo
+    VK_ICD_FILENAMES=<path-to>/MoltenVK/Package/Latest/MoltenVK/macOS/MoltenVK_icd.json vulkaninfo/vulkaninfo
```
karl-lunarg: I know we discussed this offline, but if the old docs here are correct, there was no RPATH set in the vulkaninfo executable in the install location and we were specifying that setting DYLD_LIBRARY_PATH was necessary to find the loader.  I think we need to keep this behavior.  The `../lib` relationship is much too fragile.  And we shouldn't be "making it easy" to run things out of the repo's install tree(s).  The point of the build tree is to make things easier to run (hence, the use of RPATH there), but the point of the install tree is to have binaries free of RPATH entries that may have absolute or relative filesystem references.

cube/CMakeLists.txt
```diff
@@ -200,8 +200,17 @@ else()
 endif()
 
 if(APPLE)
-    set_target_properties(cube PROPERTIES INSTALL_RPATH "@loader_path/../lib")
-    install(TARGETS cube DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+    # Keep RPATH so fixup_bundle can use it to find libraries
+    set_target_properties(cube PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE)
```
karl-lunarg: So (new) line 204 is just the "right way" to accomplish was was in (old) line 203?  (Along with calling fixup_bundle below)

mikes-lunarg: It seems that the old line 203/204 would re-add an rpath of ../lib after the fixup_bundle from cube.cmake stripped the original rpath.  This doesn't actually end up doing anything because fixup_bundle also stops the application from loading libvulkan via rpath.

BUILD.md
mikes-lunarg: I removed INSTALL_RPATH from the vulkaninfo target and changed BUILD.md back to the original wording about RPATH being removed after "make install"

PR #53, opened 2018-07-19
cmake: Remove deprecated GLSLANG_REPO_ROOT

mikew-lunarg:
The *_REPO_ROOT mechanism for finding packages is deprecated.

Passes internal CI.


mikew-lunarg:
This seems to be leftover from recent repo split.

There still exists MOLTENVK_REPO_ROOT, but that does seem to be active and entrenched code.

PR #61, opened 2018-08-22
repo: Update known-good for Vulkan-Headers

joey-lunarg:
Change-Id: I1fa27ee2564cf676418be543d6e2e2cb7f1cd906

### DIFF
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -6,7 +6,7 @@
       "sub_dir" : "glslang",
       "build_dir" : "glslang/build",
       "install_dir" : "glslang/build/install",
-      "commit" : "e99a26810f65314183163c07664a40e05647c15f",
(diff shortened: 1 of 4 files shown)


scripts/known_good.json
```diff
@@ -17,7 +17,7 @@
       "sub_dir" : "Vulkan-Headers",
       "build_dir" : "Vulkan-Headers/build",
       "install_dir" : "Vulkan-Headers/build/install",
-      "commit" : "c4e056d365472174471a243dfefbfe66a03564af"
+      "commit" : "db09f95ac00e44149f3894bf82c918e58277cfdb"
```
shannon-lunarg: I believe the Loader commit ID should be updated as well. I'm not sure about the glslang commit ID--for VVL I update it in the general scripts directory, but not for the Android known-good.

joey-lunarg: Cool, I went ahead and did them all.

PR #63, opened 2018-08-23
repo: Layer json version automation

joey-lunarg:
Change-Id: I0f71808bf58997e58599b3cb00bb23b8dd179f3c

### DIFF
--- a/cmake/FindVulkanHeaders.cmake
+++ b/cmake/FindVulkanHeaders.cmake
@@ -32,6 +32,13 @@
 #   VulkanRegistry_FOUND         - True if VulkanRegistry was found
 #   VulkanRegistry_DIRS          - directories for VulkanRegistry
 #
+#   VulkanHeaders_VERSION_MAJOR  - The Major API version of the latest version
+#                                  contained in the Vulkan header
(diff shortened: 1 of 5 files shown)


mikew-lunarg:
Are the changes to this FindVulkanHeaders.cmake made by copying over that file from another Khr repo?

If so, it would be good to identify that in this PR's comments (which repo, and at what commitID)

If not, what is the reason for deviation?  I would hope all instances of that .cmake file are identical across repos.

joey-lunarg:
These changes are copied over from Vulkan-Loader, so we already have an inconsistency. I can make the effort to make them consistent across the repos.

jeremyk-lunarg:
@shannon-lunarg @mikew-lunarg Is this still relevant?

mikew-lunarg:
Shannon did give thumbs up way back when, but it wasn't merged.
Don't know whether it's still appropriate for the current state of the tree.

jeremyk-lunarg:
Closing for now since there are merge conflicts and the tree has moved forward.

If we want something like this in the future, we can either reopen with fixes or open a new PR with necessary changes.

PR #79, opened 2018-10-19
mock_icd: fix vkGetEventStatus output

ShabbyX:
VK_SUCCESS is not a valid return value for this function.  VK_EVENT_SET
is used instead.


ShabbyX:
Thanks Tobin!

PR #86, opened 2018-11-01
cmake: Fix build crashing on make install

jeremyk-lunarg:
Fixed make install crashing when running
fixup_bundle on MacOS.

Note: Fixup_bundle crashes when the Vulkan-Loader repo is placed within the Vulkan-Tools tree, but seems to work fine when the Vulkan-Loader repo is placed outside the Vulkan-Tools Tree.


CMakeLists.txt
```diff
@@ -85,7 +85,9 @@ endif()
 message(STATUS "Using find_package to locate Vulkan")
 find_package(Vulkan)
 find_package(VulkanHeaders)
+get_filename_component(Vulkan_LIBRARY_DIR ${Vulkan_LIBRARY} DIRECTORY)
 message(STATUS "Vulkan FOUND = ${Vulkan_FOUND}")
+message(STATUS "Vulkan Lib Dir = ${Vulkan_LIBRARY_DIR}")
```
karl-lunarg: Technically not needed since the path (dir) would show up in the following line.  But I have no problem leaving this here.

cube/CMakeLists.txt
```diff
@@ -199,7 +199,7 @@ if(APPLE)
     # Fix up the library references to be self-contained within the bundle.
     install(CODE "
         include(BundleUtilities)
-        fixup_bundle(\${CMAKE_INSTALL_PREFIX}/cube/cube.app \"\" \"\")
+        fixup_bundle(\${CMAKE_INSTALL_PREFIX}/cube/cube.app \"\" ${Vulkan_LIBRARY_DIR})
```
karl-lunarg: I think you need to keep the (escaped) double-quotes here.  If the path contains a space, the path may get interpreted as two arguments by `fixup_bundle()`.   I know that CMake usually handles this pretty well in most cases, but here, under the `CODE` property and the need to escape the quotes, I'm not so sure.

You can check this pretty easily by creating a dir called `/tmp/xxx xxx`, copying your entire build tree there, and then re-run CMake and rebuild.  But I wouldn't bother and just add the quotes back.

karl-lunarg:
It might also be helpful to change the commit message to say that fixup_bundle fails when the Vulkan-Loader repo is placed within the Vulkan-Tools tree but works OK when Vulkan-Loader is placed outside of the Vulkan-Tools tree.  This may be some undetermined quirk with fixup_bundle.

This will help us remember the reason for the change and that fixup_bundle is working oddly here.

PR #88, opened 2018-11-02
Rename cube and cubepp exes to vkcube and vkcubepp

Tony-LunarG:
Move to a more consistent naming scheme for Vulkan apps.

### DIFF
--- a/cube/macOS/cube/cube.cmake
+++ b/cube/macOS/cube/cube.cmake
@@ -15,7 +15,7 @@
 # limitations under the License.
 # ~~~
 
-# Cube Application Bundle
+# VkCube Application Bundle
 
 set(cube_SRCS
     ${CMAKE_CURRENT_SOURCE_DIR}/macOS/cube/main.m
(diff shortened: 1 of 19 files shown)


Tony-LunarG:
I've created issue #90 to address the directory name issue that will need to be synchronized with a change to the Vulkan spec, which points to the cube source code as reference code.

cube/macOS/cube/cube.cmake
```diff
@@ -48,16 +48,16 @@ if(NOT ${CMAKE_GENERATOR} MATCHES "^Xcode.*")
                        COMMENT "Compiling storyboard")
 endif()
 
-add_dependencies(cube MoltenVK_icd-staging-json)
+add_dependencies(vkcube MoltenVK_icd-staging-json)
 
 # Include demo source code dir because the MacOS cube's Objective-C source includes the "original" cube application C source code.
 # Also include the MoltenVK helper files.
 target_include_directories(cube PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${MOLTENVK_DIR}/MoltenVK/include)
```
jeremyk-lunarg: cube -> vkcube needed here.

PR #91, opened 2018-11-08
Update wrappers and remove last MIR references

Tony-LunarG:


whydoubt:
It seems like two distinct actions are occurring in this one commit.  I would suggest splitting out the "Update wrappers" from the "remove last MIR references", with one commit for each.

PR #93, opened 2018-11-12
build: Update MoltenVK commit and CMake files

jeremyk-lunarg:
Update MoltenVK commit ID in known_good.json
and modified CMake files for cube, cubepp, and
vulkaninfo to accomidate the new file structure
of the MoltenVK repo.

Note: Updated Vulkan-Loader in known_good.json to pick up a fix for a problem where MacOS bundles would crash at the loader level.


jeremyk-lunarg:
This is passing CI.  If it doesn't get a review by the end of the day I'll assume it's good and merge.

karl-lunarg:
A bit late, but LGTM.

PR #115, opened 2018-12-19
vulkaninfo: don't query unsupported formats

lunarpapillo:
If vulkaninfo is compiled against a 1.1 implementation, but
being executed against a 1.0 implementation, it must not
query formats that aren't supported (or risk a driver crash).

- moved instance version information into the AppInstance struct
- set instance version information in AppCreateInstance()
- encode formats in an extensible struct for queries


jeremyk-lunarg:
Looks like it has some merge conflicts.

lunarpapillo:
I managed to get my repositories screwed up trying to set up for rebase.  I'm throwing this one out and will try again.

PR #116, opened 2018-12-20
vulkaninfo: only query supported formats

lunarpapillo:
vulkaninfo must not query formats that are not supported on a
given instance, as this provokes undefined behavior (it could
appear to work, or it could segfault in the driver).

These changes ensure that a format is only queried if it
is supported in either the base Vulkan instance or, if it is an
extension format, if that extension is supported by the Vulkan
instance.

- Moved instance version information to the AppInstance struct
  (initialized by AppCreateInstance()), which makes sense and
  allows lower-level code to determine whether a format is
  supported on the current instance version
- Formats are now encoded with their owning Vulkan version
  and extension name, so they can be checked for validity
  before making a query.

Note that this change can cause behavior changes; instances
that didn't crash on unsupported formats (and thus appeared to
support more formats than they actually did) will no longer
report those formats.  Instances that would segfault when
an unsupported formatn was queried should no longer crash.

### DIFF
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -2009,7 +2009,7 @@ static void AppDevDump(const struct AppGpu *gpu, FILE *out) {
     }
 
     bool first_in_list = true;   // Used for commas in json output
-    for (unsigned int i = 0; i < sizeof(supported_format_ranges)/sizeof(supported_format_ranges[0]); i++) {
+    for (uint32_t i = 0; i < sizeof(supported_format_ranges)/sizeof(supported_format_ranges[0]); i++) {
         struct FormatRange format_range = supported_format_ranges[i];
         if (FormatRangeSupported(&format_range, gpu)) {
             for (VkFormat fmt = format_range.first_format; fmt <= format_range.last_format; ++fmt) {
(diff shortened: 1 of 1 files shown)


vulkaninfo/vulkaninfo.c
```diff
@@ -1935,15 +2009,13 @@ static void AppDevDump(const struct AppGpu *gpu, FILE *out) {
...
-    for (VkFormat fmt = 0; fmt < VK_FORMAT_RANGE_SIZE; ++fmt) {
-        AppDevDumpFormatProps(gpu, fmt, &first_in_list, out);
-    }
-
-    for (VkFormat fmt = VK_FORMAT_G8B8G8R8_422_UNORM; fmt <= VK_FORMAT_G16_B16_R16_3PLANE_444_UNORM; ++fmt) {
-        AppDevDumpFormatProps(gpu, fmt, &first_in_list, out);
-    }
-    for (VkFormat fmt = VK_FORMAT_PVRTC1_2BPP_UNORM_BLOCK_IMG; fmt <= VK_FORMAT_PVRTC2_4BPP_SRGB_BLOCK_IMG; ++fmt) {
-        AppDevDumpFormatProps(gpu, fmt, &first_in_list, out);
+    for (unsigned int i = 0; i < sizeof(supported_format_ranges)/sizeof(supported_format_ranges[0]); i++) {
```
jeremyk-lunarg: A bit of a nit pick.  You might want to change the `unsigned int` here to a `uint32_t` for consistency with the rest of the file.

lunarpapillo: I love nit picks - it shows the reviewer is being thorough. :-)  Thanks for the review...

jeremyk-lunarg:
Changes look good.  Feel free to merge whenever you want.

PR #126, opened 2019-01-22
vulkaninfo: Make queue families report present support

### LINKED ISSUE #102: Vulkaninfo could list which queues support present
When vulkaninfo is listing the capabilities of each of the queue families (GRAPHICS, COMPUTE, etc), it would be nice if it also listed whether the queue family supports present. A call to GetPhysicalDeviceSurfaceSupportKHR is all it would take.

krOoze:
closes #102


jeremyk-lunarg:
Had to revert this change in SDK 1.1.101.0 because it uncovered issue #136 on Ubuntu systems using the Mesa Intel driver.  We are still looking into what is the root cause of the problem.

PR #135, opened 2019-02-06
vulkaninfo: Remove vkCreateInstance pNext debug callback

krOoze:
which probably does nothing; we are not even enabling any layers, and we do not even create regular debug callback later.

### DIFF
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -199,35 +199,6 @@ struct AppGpu {
     VkExtensionProperties *device_extensions;
 };
 
-// return most severe flag only
-static const char *DebugReportFlagString(const VkDebugReportFlagsEXT flags) {
-    switch (flags) {
-        case VK_DEBUG_REPORT_ERROR_BIT_EXT:
-            return "ERROR";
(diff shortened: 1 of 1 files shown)


jeremyk-lunarg:
We recently used this feature to help debug a vulkaninfo issue.  Would rather leave it in.  Will leave this open until tomorrow for further comments if desired.

krOoze:
@jeremyk-lunarg May I ask how. `pNext` version coverage is minimal, and last time I checked most of the validation check do not work anyway.

jeremyk-lunarg:
We explicitly activated validation layers by setting the VK_INSTANCE_LAYERS environment variable and ran vulkaninfo with this variable set.  This caught a validation error in the vulkaninfo code.

krOoze:
@jeremyk-lunarg In that case you do not need it because that variable should implicitly register its own callback?

jeremyk-lunarg:
Tested this.  You are right.  Thanks for the catch.  Will keep this open until @lenny-lunarg has a chance to comment.

charles-lunarg:
The Debug Report is vital for returning Loader debug messages, and as such should remain in vulkaninfo. Also, the refactor is finished and so this PR can't be merged anyhow ( as it deleted vulkaninfo.c)

PR #137, opened 2019-02-11
build: Revert known-good to 1.1.97 header

shannon-lunarg:
This commit consists of reverting known-good files to 'v1.1.97' tags due to a spec bug introduced in the 1.1.99 header and a Vulkan-Hpp failure with the 1.1.100 header.

Note - I ran this through Gerrit and it passed.


scripts/kvt_genvk.py
```diff
@@ -21,14 +21,12 @@
 
 def startTimer(timeit):
     global startTime
-    if timeit:
-        startTime = time.process_time()
+    startTime = time.clock()
```
jzulauf-lunarg: This seems like an unrelated change, is it needed or should we have a custom revert commit that only changes the known-good commit

shannon-lunarg: These changes were made during the known-good update--changes in Vulkan-Headers' `genvk.py` are reflected in each repo's version of the same file when known-good is updated-- so I reverted the changes to `kvt_genvk.py` as well. I can do some investigating to see if I need to revert `kvt_genvk.py` or if I can leave it as it.

shannon-lunarg: My conclusion is the `kvt_genvk.py` changes do not need to be reverted. These updated changes have passed through Gerrit and github CI.

PR #142, opened 2019-02-15
Fix the build error on Android

renjiayuan1314:
I know that there is no planned work for vulkaninfo for Android.
But someone will use the vulkaninfo crossing all the platforms
including Android.

This change fixed the compiling error on Android platform.

### DIFF
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -163,7 +163,7 @@ struct AppInstance {
     Window xlib_window;
 #endif
 #ifdef VK_USE_PLATFORM_ANDROID_KHR  // TODO
-    ANativeWindow *window;
+    struct ANativeWindow *window;
 #endif
 #ifdef VK_USE_PLATFORM_MACOS_MVK
     void *window;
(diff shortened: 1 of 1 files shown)


renjiayuan1314:
Just updated the commit message, please kindly review.
Thanks a lot.

jeremyk-lunarg:
How have you been building vulkaninfo for Android?  I've been looking around our repo and test systems and it doesn't look like we provide a way to do this.

renjiayuan1314:
>> How have you been building vulkaninfo for Android?

Yes, I have verified on my local Android P build. I defined the VK_USE_PLATFORM_ANDROID_KHR  in my own makefile and it passes with this change.

jeremyk-lunarg:
Yes, but what method have you been using to build it for your local Android when we do not provide functionality to build vulkaninfo on Android.

renjiayuan1314:
We are doing Android OS and driver development and build the vulkaninfo in the Android P source tree.

we imported the entire Vulkan-Tools git repository. And then we added the makefiles for Android for vulkaninfo (we didn't use other components like cube). The output is a vulkaninfo executable binary, not apk and we use it for internal testing (printing the vulkan extensions and capabilities, etc).

Hope that answers your question.

renjiayuan1314:
Thanks, Jeremy.

PR #148, opened 2019-02-25
Fixed undefined reference to 'clock_gettime' by linking rt library

francoisbertelatschrodinger:
On CentOS6.8 with a local build of g++ 6.4.0 (the default g++ version on CentOS6.8 is 4.4.7), the build fails with undefined reference to 'clock_gettime'.

Linking vkcube against the rt library fixed the issue.

This patch is similar to https://github.com/KhronosGroup/SPIRV-Tools/pull/2409


Tony-LunarG:
We have submitted #149 to attempt to address this a little more conservatively

jeremyk-lunarg:
PR#149 has reportedly fixed this problem.  Closing PR.

PR #149, opened 2019-02-26
Add librt when necessary

Tony-LunarG:
This attempts to be a little more conservative than #148  francoisbertelatschrodinger could you see if this fixes your problem?


cube/CMakeLists.txt
```diff
@@ -174,6 +174,12 @@ elseif(NOT WIN32)
                        cube.vert.inc
                        cube.frag.inc)
         target_link_libraries(vkcube Vulkan::Vulkan)
+        if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
```
mikew-lunarg: Is this CMAKE_SYSTEM_NAME check actually required?  By the outer if-else tree, it's already !APPLE and !WIN32.

Tony-LunarG: Ok, removed

francoisbertelatschrodinger:
Yes this patch fixes my problem.

Issue #136, opened 2019-02-09
Vullkaninfo broken in Linux- Segmentation fault 


arunsivaramanneo:
[vulkaninfo_Segmentation_fault.txt](https://github.com/KhronosGroup/Vulkan-Tools/files/2847507/vulkaninfo_Segmentation_fault.txt)

karl-lunarg:
It would be really helpful to run this under a debugger and post a stack trace at the point of the fault.

arunsivaramanneo:
@karl-lunarg , i am trying to debug.. but unfortunately the following command is unable to complete without any errors.. cmake -DCMAKE_BUILD_TYPE=Debug .. 

Attaching the cMakeOutputlog, 
[CMakeOutput.log](https://github.com/KhronosGroup/Vulkan-Tools/files/2848636/CMakeOutput.log)

Note : I did report this issue few weeks back .. unfortunately i was not able to give more information

https://github.com/KhronosGroup/Vulkan-Tools/commit/01402386ecc9080f0f8e0c4e194958a0868d1027

arunsivaramanneo:
@karl-lunarg .. this is the last commit thats working fine on my linux machine

https://github.com/KhronosGroup/Vulkan-Tools/commit/e1d6a0be32144be2b378532f4183d679821234c8

karl-lunarg:
It isn't necessary to build it debuggable to get a stack trace if the trace is being used just to see where the fault occurred.  See what you did [here](https://github.com/KhronosGroup/Vulkan-Tools/commit/01402386ecc9080f0f8e0c4e194958a0868d1027#commitcomment-32041643).

I suspect that this is the same problem being discussed in this [commit](https://github.com/KhronosGroup/Vulkan-Tools/commit/01402386ecc9080f0f8e0c4e194958a0868d1027).   And I agree with the assessment in this [comment](https://github.com/KhronosGroup/Vulkan-Tools/commit/01402386ecc9080f0f8e0c4e194958a0868d1027#commitcomment-32042292):

> I'm asking because I notice that commit [3a7c6a8](https://github.com/KhronosGroup/Vulkan-Tools/commit/3a7c6a8413130d146492c688893305ff2652311a) (two commits ago) introduced the call to `vkGetPhysicalDeviceSurfaceSupportKHR` that seems to appear in the crash backtrace.

[...]

PR #153, opened 2019-03-05
vulkaninfo: Fix window\surface lifetime

krOoze:


jeremyk-lunarg:
Looks like there are some merge conflicts.
Also, remember to clang-format your code.

krOoze:
> Also, remember to clang-format your code.

DAMMIT, not again!

Yea the offending commit seems to do something similar to this. It does not seem to manage window lifetime though, which I assumed would be the problem; does it actually work to fix #136 (I am still waiting on confirmation for this PR too)? I may be biased but my commit feels cleaner managing the lifetime on the same scope. Not sure what to do except revert that commit if that won't offend you...

jeremyk-lunarg:
Clang-format gets me a lot too.  In fact, I have a piece of paper taped to my monitor to remind me.

It won't offended me to revert the commit.  Your implementation looks cleaner.  I am a bit curious why you are using an array instead of a linked list to handle multiple surfaces.

krOoze:
I am not a natural C programmer, so I welcome convention notes.
In C++ I would use `std::vector`. So I do the same here the closest I can approximate it...

krOoze:
Not sure what the CI complains about now. Seems glitched. Would you PTAL?

vulkaninfo/vulkaninfo.c
```diff
@@ -4542,24 +4517,19 @@ static void AppGpuDumpQueueProps(const struct AppGpu *gpu, uint32_t id, FILE *ou
                 "class='val'>%d</div>, <div class='val'>%d</div>)</summary></details>\n",
                 props.minImageTransferGranularity.width, props.minImageTransferGranularity.height,
                 props.minImageTransferGranularity.depth);
-        fprintf(out, "\t\t\t\t\t\t<details><summary>present support</summary>\n");
-        for (struct SurfaceNameWHandle *surface_stuff = gpu->inst->surface_chain; surface_stuff != NULL;
-             surface_stuff = surface_stuff->pNextSurface) {
-            fprintf(out, "\t\t\t\t\t\t\t<details><summary>%s = <div class='val'>%s</div></summary></details>\n",
-                    surface_stuff->name, surface_stuff->present_support ? "true" : "false");
+        for (int i = 0; i < gpu->inst->surface_ext_infos_count; ++i) {
+            fprintf(out, "\t\t\t\t\t\t<details><summary>present support(%s) = <div class='val'>%s</div></summary></details>\n",
```
jeremyk-lunarg: Would prefer if this was made into a section that listed present support for each surface type under it.
So instead of:
```
VkQueueFamilyProperties[0]:
===========================
        queueFlags         = GRAPHICS | COMPUTE | TRANSFER | SPARSE
        queueCount         = 16
        timestampValidBits = 64
        minImageTransferGranularity = (1, 1, 1)
        present support(VK_KHR_xcb_surface) = true
        present support(VK_KHR_xlib_surface) = true
```
It would be:
```
VkQueueFamilyProperties[0]:
===========================
        queueFlags         = GRAPHICS | COMPUTE | TRANSFER | SPARSE
        queueCount         = 16
        timestampValidBits = 64
        minImageTransferGranularity = (1, 1, 1)
        present support:
                VK_KHR_xcb_surface = true
                VK_KHR_xlib_surface = false
```

vulkaninfo/vulkaninfo.c
```diff
@@ -4542,24 +4517,19 @@ static void AppGpuDumpQueueProps(const struct AppGpu *gpu, uint32_t id, FILE *ou
...
         printf("\tqueueCount         = %u\n", props.queueCount);
         printf("\ttimestampValidBits = %u\n", props.timestampValidBits);
         printf("\tminImageTransferGranularity = (%d, %d, %d)\n", props.minImageTransferGranularity.width,
                props.minImageTransferGranularity.height, props.minImageTransferGranularity.depth);
-        printf("\tpresent support:\n");
-        for (struct SurfaceNameWHandle *surface_stuff = gpu->inst->surface_chain; surface_stuff != NULL;
-             surface_stuff = surface_stuff->pNextSurface) {
-            printf("\t\t%s = %s\n", surface_stuff->name, surface_stuff->present_support ? "true" : "false");
+        for (int i = 0; i < gpu->inst->surface_ext_infos_count; ++i) {
+            printf("\tpresent support(%s) = %s\n", gpu->inst->surface_ext_infos[i].name, supports_present ? "true" : "false");
```
jeremyk-lunarg: See above comment.

vulkaninfo/vulkaninfo.c
```diff
@@ -4572,16 +4542,10 @@ static void AppGpuDumpQueueProps(const struct AppGpu *gpu, uint32_t id, FILE *ou
...
-        for (struct SurfaceNameWHandle *surface_stuff = gpu->inst->surface_chain; surface_stuff != NULL;
-             surface_stuff = surface_stuff->pNextSurface) {
-            if (surface_stuff->pNextSurface) {
-                printf("\t\t\t\t\"%s\" : %u,\n", surface_stuff->name, surface_stuff->present_support);
-            } else {
-                printf("\t\t\t\t\"%s\" : %u\n", surface_stuff->name, surface_stuff->present_support);
-            }
+        for (int i = 0; i < gpu->inst->surface_ext_infos_count; ++i) {
+            printf("\t\t\t\"present_support(%s)\": %s\n", gpu->inst->surface_ext_infos[i].name,
+                   supports_present ? "\"true\"" : "\"false\"");
```
jeremyk-lunarg: See above comment.

vulkaninfo/vulkaninfo.c
```diff
@@ -4902,13 +4866,8 @@ static void AppGroupDump(const VkPhysicalDeviceGroupProperties *group, const uin
                                               .physicalDeviceCount = group->physicalDeviceCount,
                                               .pPhysicalDevices = group->physicalDevices};
 
-    float queue_priority = 1.0f;
```
jeremyk-lunarg: This was added to fix a validation error.
I can add this back later, but if you'd like you can add this back in for this PR.

vulkaninfo/vulkaninfo.c
```diff
@@ -5407,7 +5384,11 @@ int main(int argc, char **argv) {
     free(gpus);
     free(objs);
 
-    AppDestroySurfaceChain(&inst);
+    for (int i = 0; i < inst.surface_ext_infos_count; ++i) {
```
jeremyk-lunarg: Once this loop is done, you would have to free `inst.surface_ext_infos` because it is heap memory.

vulkaninfo/vulkaninfo.c
```diff
@@ -897,9 +892,14 @@ static void AppCreateInstance(struct AppInstance *inst) {
     }
 
     AppLoadInstanceCommands(inst);
+
+    inst->surface_ext_infos_reserve = 10;
```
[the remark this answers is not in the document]
jeremyk-lunarg: Also, there are only 5 possible surfaces that vulkaninfo supports.  So if you choose to keep this, you don't need to reserve memory for 10 `struct SurfaceExtensionInfo`.

vulkaninfo/vulkaninfo.c
krOoze: I am freeing it in Instance destructor.

vulkaninfo/vulkaninfo.c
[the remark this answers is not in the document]
krOoze: There are 9 platforms I think, though most of them won't coexist. Some crazy mock driver could implement all of them though. And some platforms will probably be added in the future.

vulkaninfo/vulkaninfo.c
```diff
@@ -112,11 +112,12 @@ struct LayerExtensionList {
     VkExtensionProperties *extension_properties;
 };
 
-struct SurfaceNameWHandle {
+struct SurfaceExtensionInfo {
```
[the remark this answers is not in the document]
krOoze: Maybe. We only need iterable container anyway.

vulkaninfo/vulkaninfo.c
krOoze: I suppose.
One other thing: I wasn't sure how to inform this is the platform-agnostic query. Should something be done there?

krOoze: Though I think perhaps the html and human readable output could show only one value assuming it is same for all the WSI platforms....

vulkaninfo/vulkaninfo.c
krOoze: OK, anything other worth preserving?

jeremyk-lunarg: Lines 4905-4911 should be all you need.  The rest of the code was similar to what you are doing in this commit.

vulkaninfo/vulkaninfo.c
```diff
@@ -4572,16 +4558,11 @@ static void AppGpuDumpQueueProps(const struct AppGpu *gpu, uint32_t id, FILE *ou
...
-        for (struct SurfaceNameWHandle *surface_stuff = gpu->inst->surface_chain; surface_stuff != NULL;
-             surface_stuff = surface_stuff->pNextSurface) {
-            if (surface_stuff->pNextSurface) {
-                printf("\t\t\t\t\"%s\" : %u,\n", surface_stuff->name, surface_stuff->present_support);
-            } else {
-                printf("\t\t\t\t\"%s\" : %u\n", surface_stuff->name, surface_stuff->present_support);
-            }
+        printf("\t\t\t\"present_support\": [\n");
+        for (struct SurfaceExtensionNode *sen = gpu->inst->surface_ext_infos_root; sen != NULL; sen = sen->next) {
+            printf("\t\t\t\t\"%s\": \"%s\"%s\n", sen->name, sen->supports_present ? "true" : "false", sen->next ? "," : "");
```
jeremyk-lunarg: @mikew-lunarg can correct me on this, but shouldn't this to be a 1 or 0 instead of a "true" or "false" string since it is JSON output?

krOoze: json has bool this way. Not sure about conventions in this repo about it...

mikew-lunarg: @jeremyk-lunarg @krOoze  vulkaninfo's JSON was specifically designed to be consumed by the device simulation (devsim) layer.   The [devsim JSON schema](https://schema.khronos.org/vulkan/devsim_1_0_0.json) defines its types directly from vulkan_core.h so Vulkan's booleans, which are VkBool32 and derived from uint32_t, should be integer values.

If there isn't already, there should be a comment in vulkaninfo that the generated JSON must validate against the devsim schema at https://schema.khronos.org/vulkan/devsim_1_0_0.json

krOoze: That's fine with me. Features seem to be binary numbers too...

mikew-lunarg:
@jeremyk-lunarg  It would be good to run some sample output thru JSON validation.  I can show you how to do that if needed.

Unfortunately, I've not found an automatable tool that we could put in CI; the validators I've found are all websites into which you paste text and press a button.  If anyone has recommendations for automatable tools, I'd welcome feedback.

PR #158, opened 2019-03-11
vulkaninfo: Report images compatible with memory type

krOoze:
This reports for Memory Types which Images they are usable with (via `vkGetImageMemoryRequirements`).
Currently it works only for images, and outputs only the `human_readable` text.

I want to solicit some input:
1) What would be the best (better) output format
2) What would be the best json (for device simulation)
3) There's bit of a combinatorial explosion of `usage`x`flags` for buffers. Any ideas how to handle that and present it to the user?

### DIFF
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -1112,14 +1112,14 @@ static void AppGpuInit(struct AppGpu *gpu, struct AppInstance *inst, uint32_t id
     if (err) ERR_EXIT(err);
 
     const VkFormat color_format = VK_FORMAT_R8G8B8A8_UNORM;
-    const VkImageTiling formats[] = {
+    const VkFormat formats[] = {
         color_format,      VK_FORMAT_D16_UNORM,         VK_FORMAT_X8_D24_UNORM_PACK32, VK_FORMAT_D32_SFLOAT,
         VK_FORMAT_S8_UINT, VK_FORMAT_D16_UNORM_S8_UINT, VK_FORMAT_D24_UNORM_S8_UINT,   VK_FORMAT_D32_SFLOAT_S8_UINT};
     assert(ARRAY_SIZE(gpu->mem_type_res_support.image[0]) == ARRAY_SIZE(formats));
     const VkImageUsageFlags usages[] = {0, VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT};
     const VkImageCreateFlags flagss[] = {0, VK_IMAGE_CREATE_SPARSE_BINDING_BIT};
 
-    for (int fmt_i = 0; fmt_i < ARRAY_SIZE(formats); ++fmt_i) {
+    for (size_t fmt_i = 0; fmt_i < ARRAY_SIZE(formats); ++fmt_i) {
         for (VkImageTiling tiling = VK_IMAGE_TILING_OPTIMAL; tiling <= VK_IMAGE_TILING_LINEAR; ++tiling) {
             gpu->mem_type_res_support.image[tiling][fmt_i].format = formats[fmt_i];
             gpu->mem_type_res_support.image[tiling][fmt_i].regular_supported = true;
@@ -1136,8 +1136,8 @@ static void AppGpuInit(struct AppGpu *gpu, struct AppInstance *inst, uint32_t id
                 continue;
             }
 
-            for (int u_i = 0; u_i < ARRAY_SIZE(usages); ++u_i) {
-                for (int flg_i = 0; flg_i < ARRAY_SIZE(flagss); ++flg_i) {
+            for (size_t u_i = 0; u_i < ARRAY_SIZE(usages); ++u_i) {
+                for (size_t flg_i = 0; flg_i < ARRAY_SIZE(flagss); ++flg_i) {
                     VkImageCreateInfo image_ci = {
                         .sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO,
                         .flags = flagss[flg_i],
@@ -1157,7 +1157,7 @@ static void AppGpuInit(struct AppGpu *gpu, struct AppInstance *inst, uint32_t id
                     }
 
                     if (image_ci.usage == 0 || (image_ci.usage & VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT)) {
-                        if (image_ci.format = color_format)
+                        if (image_ci.format == color_format)
                             image_ci.usage |= VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
                         else
                             image_ci.usage |= VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT;
@@ -4907,7 +4907,7 @@ static void AppGpuDumpMemoryProps(const struct AppGpu *gpu, FILE *out) {
                 printf("\t\t\t%s: ", VkTilingString(tiling));
 
                 bool first = true;
-                for (int fmt_i = 0; fmt_i < ARRAY_SIZE(gpu->mem_type_res_support.image[tiling]); ++fmt_i) {
+                for (size_t fmt_i = 0; fmt_i < ARRAY_SIZE(gpu->mem_type_res_support.image[tiling]); ++fmt_i) {
                     const struct MemImageSupport *image_support = &gpu->mem_type_res_support.image[tiling][fmt_i];
                     const bool regular_compatible =
                         image_support->regular_supported && (image_support->regular_memtypes & memtype_bit);
@@ -4958,7 +4958,7 @@ static void AppGpuDumpMemoryProps(const struct AppGpu *gpu, FILE *out) {
                 fprintf(out, "\t\t\t\t\t\t\t\t\t<details><summary>%s</summary>\n", VkTilingString(tiling));
 
                 bool first = true;
-                for (int fmt_i = 0; fmt_i < ARRAY_SIZE(gpu->mem_type_res_support.image[tiling]); ++fmt_i) {
+                for (size_t fmt_i = 0; fmt_i < ARRAY_SIZE(gpu->mem_type_res_support.image[tiling]); ++fmt_i) {
                     const struct MemImageSupport *image_support = &gpu->mem_type_res_support.image[tiling][fmt_i];
                     const bool regular_compatible =
                         image_support->regular_supported && (image_support->regular_memtypes & memtype_bit);
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -183,6 +183,15 @@ struct AppInstance {
 #endif
 };
 
+struct MemResSupport {
+    struct MemImageSupport {
+        bool regular_supported, sparse_supported, transient_supported;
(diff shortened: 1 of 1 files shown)


jeremyk-lunarg:
A few general suggestions:

1. I like to put a `:` instead of an `=` at the end of a list title for human readable output.
2. When you are listing multiple values (with a few exceptions when listing bits set in a bit-flag variable) it is preferred to separate them with a `\n` instead of a `,`.
3. When there is nothing to list, it is nice to let the user know explicitly by listing 'None' in the list.

So, for example:

```
memoryTypes[0] :
        heapIndex     = 1
        propertyFlags = 0x0:
        usable for =
                OPTIMAL:
                LINEAR:
memoryTypes[1] :
        heapIndex     = 1
        propertyFlags = 0x0:
        usable for =
                OPTIMAL: color images, D16_UNORM, X8_D24_UNORM_PACK32, D32_SFLOAT, D24_UNORM_S8_UINT, D32_SFLOAT_S8_UINT
                LINEAR:
```

Would become:

```
memoryTypes[0] :
        heapIndex     = 1
        propertyFlags = 0x0:
        usable for:
               OPTIMAL:
                    None
                LINEAR:
                    None
memoryTypes[1] :
        heapIndex     = 1
        propertyFlags = 0x0:
        usable for:
                OPTIMAL:
                    color images
                    D16_UNORM
                    X8_D24_UNORM_PACK32
                    D32_SFLOAT
                    D24_UNORM_S8_UINT
                    D32_SFLOAT_S8_UINT
                LINEAR:
                    None
```

As for suggestions for JSON stuff, @mikew-lunarg recently wrote up a document for vulkaninfo JSON output [here](https://github.com/KhronosGroup/Vulkan-Tools/blob/master/vulkaninfo/json_validation_process.md).

I have to think on the buffers a bit more.

krOoze:
ad 1: Yea. Trivial change...
ad 2:  I specifically did not want to spam the console. Considering the NV has like 15 types, it may get riddiculous fast. The output is at maximum color + all the depth formats and easily fits one or two lines.
ad 3: Yea. Will require more ugly boilerplate code though... ugh

ad JSON: that like just mentions the schema. I meant more specifically what to put there. I mean the schema obviously does not contain this yet as it is added info.
The audience for the JSON is bit different than the human readable output, and probably should be more tuned for computer consumption. I am not really consumer of it, so I ask for some input before I decide the format autocratically.

ad Buffers: There's two things to consider. Brute force would be few 1000s of cases. Question is if we should even attempt it. It will probably not be that slow, as the query commands should be fast enough.
Second is how to report this. I was thinking reporting each flag in isolation. Then report odd cases (i.e. not exact intersection of results for the flags in isolation) of flag combinations (which I assume there would not be many on sane driver), with limited to say 20 cases.
I also https://github.com/KhronosGroup/Vulkan-Docs/issues/940 . It seems weird the output for buffers is less restricted than images. Might be worth waiting for that resolution.

mikew-lunarg:
@krOoze If you have specific questions re JSON formatting and DevSim, let me know, I'm glad to help.

As for comma-separated lists: agreed that for small elements (like pipelineCacheUUID), the newline-between-elements thing eats lots of vertical space, and so could be combined on a single line.

I'm mainly concerned about vulkaninfo generating valid and validated JSON for machine consumption.  Tools like [jq](https://stedolan.github.io/jq/) can be used afterwards to make it look pretty.

krOoze:
@mikew-lunarg Well, e.g. the `present_support` does apparently already violate the schema. So we can use that as an example. How am I supposed to deal with that when exposing new information? Ask for schema change?

krOoze:
@mikew-lunarg Well how am I supposed to deal with exposing new info. Ask for schema change? Or does the schema tolarate additional info?

mikew-lunarg:
@krOoze The DevSim schema only dictates the structures that DevSim cares about.  It's deliberately _not_ called the Vulkan schema.  The DevSim schema will change over time as DevSim evolves, so we'll want to ensure new data is formatted properly.

To your question : how to deal with exposing new info, it depends on where the additions are going.

The `"additionalProperties": true` on line 7 of the [DevSim schema](https://schema.khronos.org/vulkan/devsim_1_0_0.json) indicates that structures beyond those defined in the schema are permitted.  So you can add new structures without schema modification.  However, when implementing new output, please keep in mind how a schema would be defined to specify that data.

The structures that the schema _does_ specify are defined as `"additionalProperties": false` and so cannot be modified without a schema change, but should already be comprehensively defined.

The best way to learn if your output is colliding with a schema is to run the validation process, as described in the doc @jeremyk-lunarg mentioned above.

For specific JSON questions, please start a new issue for design discussion, to help keep the conversations modular. 

One other point : I'm adding extension support to DevSim, with the structures for each extension being in their own schema.  We can discuss that more in an issue as needed.

krOoze:
Added HTML output
Ignored json output
Changed to `:` (suggestion 1)
Added "None" (suggestion 3)

vulkaninfo/vulkaninfo.c
```diff
@@ -1065,9 +1093,127 @@ static void AppGpuInit(struct AppGpu *gpu, struct AppInstance *inst, uint32_t id
...
+
+    const VkFormat color_format = VK_FORMAT_R8G8B8A8_UNORM;
+    const VkImageTiling formats[] = {
+        color_format,      VK_FORMAT_D16_UNORM,         VK_FORMAT_X8_D24_UNORM_PACK32, VK_FORMAT_D32_SFLOAT,
+        VK_FORMAT_S8_UINT, VK_FORMAT_D16_UNORM_S8_UINT, VK_FORMAT_D24_UNORM_S8_UINT,   VK_FORMAT_D32_SFLOAT_S8_UINT};
+    assert(ARRAY_SIZE(gpu->mem_type_res_support.image[0]) == ARRAY_SIZE(formats));
+    const VkImageUsageFlags usages[] = {0, VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT};
+    const VkImageCreateFlags flagss[] = {0, VK_IMAGE_CREATE_SPARSE_BINDING_BIT};
+
+    for (int fmt_i = 0; fmt_i < ARRAY_SIZE(formats); ++fmt_i) {
```
jeremyk-lunarg: This causes a warning, and `uint32_t` is used in the rest of the program as opposed to `int`.

vulkaninfo/vulkaninfo.c
```diff
@@ -1065,9 +1093,127 @@ static void AppGpuInit(struct AppGpu *gpu, struct AppInstance *inst, uint32_t id
...
+            vkGetPhysicalDeviceFormatProperties(gpu->obj, formats[fmt_i], &fmt_props);
+            if ((tiling == VK_IMAGE_TILING_OPTIMAL && fmt_props.optimalTilingFeatures == 0) ||
+                (tiling == VK_IMAGE_TILING_LINEAR && fmt_props.linearTilingFeatures == 0)) {
+                gpu->mem_type_res_support.image[tiling][fmt_i].regular_supported = false;
+                gpu->mem_type_res_support.image[tiling][fmt_i].sparse_supported = false;
+                gpu->mem_type_res_support.image[tiling][fmt_i].transient_supported = false;
+                continue;
+            }
+
+            for (int u_i = 0; u_i < ARRAY_SIZE(usages); ++u_i) {
```
jeremyk-lunarg: See above comment.

vulkaninfo/vulkaninfo.c
```diff
@@ -1065,9 +1093,127 @@ static void AppGpuInit(struct AppGpu *gpu, struct AppInstance *inst, uint32_t id
...
+                for (int flg_i = 0; flg_i < ARRAY_SIZE(flagss); ++flg_i) {
```
jeremyk-lunarg: See above comment.

vulkaninfo/vulkaninfo.c
```diff
@@ -1065,9 +1093,127 @@ static void AppGpuInit(struct AppGpu *gpu, struct AppInstance *inst, uint32_t id
...
+                        .usage = usages[u_i],
+                    };
+
+                    if ((image_ci.flags & VK_IMAGE_CREATE_SPARSE_BINDING_BIT) &&
+                        (image_ci.usage & VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT)) {
+                        continue;
+                    }
+
+                    if (image_ci.usage == 0 || (image_ci.usage & VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT)) {
+                        if (image_ci.format = color_format)
```
jeremyk-lunarg: This causes a warning.  Is this supposed to be a Boolean or assignment operation?

vulkaninfo/vulkaninfo.c
```diff
@@ -4753,7 +4898,114 @@ static void AppGpuDumpMemoryProps(const struct AppGpu *gpu, FILE *out) {
...
+
+        if (human_readable_output) {
+            printf("\t\tusable for:\n");
+            const uint32_t memtype_bit = 1 << i;
+
+            for (VkImageTiling tiling = VK_IMAGE_TILING_OPTIMAL; tiling < ARRAY_SIZE(gpu->mem_type_res_support.image); ++tiling) {
+                printf("\t\t\t%s: ", VkTilingString(tiling));
+
+                bool first = true;
+                for (int fmt_i = 0; fmt_i < ARRAY_SIZE(gpu->mem_type_res_support.image[tiling]); ++fmt_i) {
```
jeremyk-lunarg: See above comments on `uint32_t` vs `int`.

vulkaninfo/vulkaninfo.c
```diff
@@ -4753,7 +4898,114 @@ static void AppGpuDumpMemoryProps(const struct AppGpu *gpu, FILE *out) {
...
+            }
+        } else if (html_output) {
+            fprintf(out, "\t\t\t\t\t\t\t\t<details><summary>usable for</summary>\n");
+            const uint32_t memtype_bit = 1 << i;
+
+            for (VkImageTiling tiling = VK_IMAGE_TILING_OPTIMAL; tiling < ARRAY_SIZE(gpu->mem_type_res_support.image); ++tiling) {
+                fprintf(out, "\t\t\t\t\t\t\t\t\t<details><summary>%s</summary>\n", VkTilingString(tiling));
+
+                bool first = true;
+                for (int fmt_i = 0; fmt_i < ARRAY_SIZE(gpu->mem_type_res_support.image[tiling]); ++fmt_i) {
```
jeremyk-lunarg: See above comment.

vulkaninfo/vulkaninfo.c
krOoze: OK, I will check the CI logs...

vulkaninfo/vulkaninfo.c
krOoze: It is supposed to be a bug.

vulkaninfo/vulkaninfo.c
krOoze: All compiler warnings should be fixed now...

krOoze:
@jeremyk-lunarg Yes, the html format is collapsible, so it is bit more verbose.

In the human readable output sparse and transient is only noted in the case they are supported by the image format in the first place and only if they are not compatible with the memory type, OR if they are compatible but the regular image type is not.

PR #160, opened 2019-03-12
vulkaninfo: Comply with json schema for present support

krOoze:
Suggested json format:

```
    "ArrayOfVkQueueFamilyProperties": [
        {},{}
    ],
    "ArrayOfQueueSurfaceSupport": [
        [
            {"VK_KHR_win32_surface": 1},
            {"VK_KHR_whatev_surface": 0}
        ],
        [
            {"VK_KHR_win32_surface": 0}
            {"VK_KHR_whatev_surface": 1}
        ]
    ],
    "VkPhysicalDeviceMemoryProperties": {
    }
```

### DIFF
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -4560,7 +4560,7 @@ static void AppGpuDumpQueueProps(const struct AppGpu *gpu, uint32_t id, FILE *ou
         printf("\t\t\t\"timestampValidBits\": %u,\n", props.timestampValidBits);
         printf("\t\t\t\"present_support\": [\n");
         for (struct SurfaceExtensionNode *sen = gpu->inst->surface_ext_infos_root; sen != NULL; sen = sen->next) {
-            printf("\t\t\t\t\"%s\": %s%s\n", sen->name, sen->supports_present ? "1" : "0", sen->next ? "," : "");
+            printf("\t\t\t\t{\"%s\": %s}%s\n", sen->name, sen->supports_present ? "1" : "0", sen->next ? "," : "");
         }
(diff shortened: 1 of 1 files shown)


mikew-lunarg:
Well, ArrayOfQueueSurfaceSupport would require definition of a QueueSurfaceSupport structure.  The current design approach has been to pull in Vk types as exactly as possible from vulkan_core.h. 
 QueueSurfaceSupport seems to be a container for pieces derived from other structures.

What's the goal here?  Are you expecting to use ArrayOfQueueSurfaceSupport in some JSON consumer?

If not, and this change is only for parity with the HTML side, I'd be more inclined to not implement a JSON representation at all.  When there is a JSON consumer requiring that data, we can then decide how best to implement.  So feel free to do HTML-only if that's your primary goal.

krOoze:
> What's the goal here? Are you expecting to use ArrayOfQueueSurfaceSupport in some JSON consumer?
>
> If not, and this change is only for parity with the HTML side, I'd be more inclined to not implement a JSON representation at all.

AIS I am not even a direct consumer of the json. So someone else should ideally make the call.

I simply assume any and all info would be usable for DevSim for it to be as accurate as possible. DevSim would need this in order to simulate `vkGetPhysicalDeviceSurfaceSupportKHR` command.

mikew-lunarg:
@krOoze Ok I'll make the call: don't worry about implementing JSON output.    @jeremyk-lunarg FYI

When a tool requires that new info, that will motivate creation of the JSON schema that defines the output formatting.

krOoze:
@mikew-lunarg Ok then, I will repurpose the other PR for removing the json output.

PR #162, opened 2019-03-15
deps: allow a --generator switch to update_deps.py

lunarpapillo:
This allows you to use any CMake generator available
on your system to configure the dependent repositories,
allowing you to match the same generator being used by
the project build.

update_deps.py:
- Take a parameter to specify a CMake generator, and pass it to CMake
  if present
- Allow the user to reduce the count of parallel make jobs that can
  happen at one time
- remove an unnecessary semicolon

### DIFF
--- a/scripts/update_deps.py
+++ b/scripts/update_deps.py
@@ -415,6 +415,12 @@ class GoodRepo(object):
                 cmake_cmd.append('-A')
                 cmake_cmd.append('x64')
 
+        # Apply a generator, if one is specified.  This can be used to supply
+        # a specific generator for the dependent repositories to match
+        # that of the main repository.
+        if self._args.generator is not None:
+            cmake_cmd.extend(['-G', self._args.generator])
+
         if VERBOSE:
             print("CMake command: " + " ".join(cmake_cmd))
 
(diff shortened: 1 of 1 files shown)


scripts/update_deps.py
```diff
@@ -435,8 +441,13 @@ def CMakeBuild(self):
...
         if platform.system() == 'Linux' or platform.system() == 'Darwin':
             cmake_cmd.append('--')
-            cmake_cmd.append('-j{ncpu}'
-                             .format(ncpu=multiprocessing.cpu_count()))
+            num_make_jobs = multiprocessing.cpu_count()
+            if 'MAKE_JOBS' in os.environ:
+                try:
+                    num_make_jobs = min(num_make_jobs, int(os.environ['MAKE_JOBS']))
+                except ValueError:
+                    pass
```
jeremyk-lunarg: Is there a reason we are catching an error only to suppress it?

lunarpapillo: Yes, though I clearly neglected to document it it comments...

I took it as an objective to ensure that these changes would not break any existing environments.  In the (IMHO unlikely, but possible) case that some environment using update_deps happened to have a non-numeric MAKE_JOBS environment variable, I didn't want to either crash or produce an unexpected warning message, so I just continued as the code would have before.

If you disagree with the logic, I'm happy to make a different response - I'm not particularly fond of silent exception handling either.

jeremyk-lunarg: I think it would be good to let the user know that their 'MAKE_JOBS' environment variable is in a weird state and tell them how many jobs we will be performing instead.

This would hopefully help the user debug their system instead of thinking that our script doesn't listen to the environment variables they have set.

lunarpapillo: Done...

lunarpapillo:
We want the same update_deps across all repos.  Adding the other repo representatives here for review so we can come to consensus on the final form of update_deps.py, and then I'll copy the changes to the other repos.

lunarpapillo:
No comments from others heard, so I'm taking this as final and merging it into other repos.

Issue #163, opened 2019-03-16
vulkaninfo html output not valid html5


krOoze:
vulkaninfo html output shows validation errors at `https://validator.w3.org`.
Notably:
1) missing charset
2)  `div` tag not allowed in `summary` tag. Should probably use `span`...

PR #166, opened 2019-03-21
 vulkaninfo: Fix HTML output warnings and errors 

jeremyk-lunarg:
Updated vulkaninfo HTML output to fix errors and
warnings output by https://validator.w3.org.

Fixes issue #163


jeremyk-lunarg:
This is passing CI.  If there is no comment by EOD I'll assume it is good and rebase and merge.

PR #172, opened 2019-04-04
Fix memory leak in `vkcube`.

Jason2013:
Allocated memory is not freed after being used. This causes memory leak.


jeremyk-lunarg:
Changes look good and it's passing on our local CI.
Global CI seems to be screwing up though.
I'll take a look at this and get back to you.

jeremyk-lunarg:
Global CI should be fixed.  Please rebase and resubmit your change so that it can run through CI.

PR #174, opened 2019-04-12
Coverity fixes

airlied:
Our internal coverity scan ran over this, and I thought I'd send some of the trivial fixes.

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -1084,6 +1084,9 @@ static void demo_draw(struct demo *demo) {
         .pImageIndices = &demo->current_buffer,
     };
 
+    VkRectLayerKHR rect;
+    VkPresentRegionKHR region;
+    VkPresentRegionsKHR regions;
     if (demo->VK_KHR_incremental_present_enabled) {
         // If using VK_KHR_incremental_present, we provide a hint of the region
         // that contains changed content relative to the previously-presented
@@ -1093,23 +1096,20 @@ static void demo_draw(struct demo *demo) {
         // ensure that the entire image has the correctly-drawn content.
         uint32_t eighthOfWidth = demo->width / 8;
         uint32_t eighthOfHeight = demo->height / 8;
-        VkRectLayerKHR rect = {
-            .offset.x = eighthOfWidth,
-            .offset.y = eighthOfHeight,
-            .extent.width = eighthOfWidth * 6,
-            .extent.height = eighthOfHeight * 6,
-            .layer = 0,
-        };
-        VkPresentRegionKHR region = {
-            .rectangleCount = 1,
-            .pRectangles = &rect,
-        };
-        VkPresentRegionsKHR regions = {
-            .sType = VK_STRUCTURE_TYPE_PRESENT_REGIONS_KHR,
-            .pNext = present.pNext,
-            .swapchainCount = present.swapchainCount,
-            .pRegions = &region,
-        };
+
+        rect.offset.x = eighthOfWidth;
+        rect.offset.y = eighthOfHeight;
+        rect.extent.width = eighthOfWidth * 6;
+        rect.extent.height = eighthOfHeight * 6;
+        rect.layer = 0;
+
+        region.rectangleCount = 1;
+        region.pRectangles = &rect;
+
+        regions.sType = VK_STRUCTURE_TYPE_PRESENT_REGIONS_KHR;
+        regions.pNext = present.pNext;
+        regions.swapchainCount = present.swapchainCount;
+        regions.pRegions = &region;
         present.pNext = &regions;
     }
 
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -2918,7 +2918,6 @@ static void demo_init_vk(struct demo *demo) {
     VkResult err;
     uint32_t instance_extension_count = 0;
     uint32_t instance_layer_count = 0;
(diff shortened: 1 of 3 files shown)


vulkaninfo/vulkaninfo.c
```diff
@@ -4716,7 +4716,7 @@ static void AppDumpExtensions(const char *indent, const char *layer_name, const
         printf("\tcount = %d\n", extension_count);
     }
 
-    const bool is_device_type = strcmp(layer_name, "Device") == 0;
+    const bool is_device_type = layer_name && (strcmp(layer_name, "Device") == 0);
```
jeremyk-lunarg: strcmp could be switched to strncmp with the limit as VK_MAX_EXTENSION_NAME_SIZE.

jeremyk-lunarg:
Thanks for doing this.

vulkaninfo/vulkaninfo.c
krOoze: Hmm the variable name is bit bad.
Anyway the function seems to assume the pointer is valid; it is never called with `NULL` and it would not make sense if it was.

Perhaps there shoud just be `assert(layer_name && strlen(layer_name) > 0)`.

vulkaninfo/vulkaninfo.c
```diff
@@ -5500,6 +5500,10 @@ int main(int argc, char **argv) {
 
     if (html_output) {
         out = fopen("vulkaninfo.html", "w");
+        if (!out) {
+            printf("Unable to open vulkaninfo.html for writing\n");
+            return 1;
```
krOoze: Not sure if this is conventional in this file.
Typically, if error happens `ERR_EXIT` is called with some fake `VkResult` code (e.g. if `malloc` fails `ERR_EXIT(VK_ERROR_OUT_OF_HOST_MEMORY)` is called).

cube/cube.c
```diff
@@ -1093,23 +1096,20 @@ static void demo_draw(struct demo *demo) {
...
+        rect.extent.height = eighthOfHeight * 6;
+        rect.layer = 0;
+
+        region.rectangleCount = 1;
+        region.pRectangles = &rect;
+
+        regions.sType = VK_STRUCTURE_TYPE_PRESENT_REGIONS_KHR;
+        regions.pNext = present.pNext;
+        regions.swapchainCount = present.swapchainCount;
+        regions.pRegions = &region;
```
krOoze: This still seems weird\brittle. `pRegions` here assumes one element, but above some external `swapchainCount` is provided.

vulkaninfo/vulkaninfo.c
airlied: The first time layer_name is used in that function, it's checked if layer_name is valid, but yes it never seems to get called with NULL, so maybe an assert is more suitable.

airlied: So this make me wonder if the calling code is always correct, since I can't infer the intent here,

below is some of the calling code, why does it pass "Layer-Device" instead of the layer_name?

        char *layer_name = inst.global_layers[i].layer_properties.layerName;
            AppGetPhysicalDeviceLayerExtensions(&gpus[j], layer_name, &count, &props);
            if (html_output) {
                AppDumpExtensions("\t\t\t", "Layer-Device", count, props, out);
            } else if (human_readable_output) {
                AppDumpExtensions("\t\t", "Layer-Device", count, props, out);
            }

vulkaninfo/vulkaninfo.c
airlied: The previous usage failures, do a return 1; also not sure we have a return code for can't open file.

cube/cube.c
airlied: It's explicitly set to 1 in the present struct just a few lines earlier, not sure it helps changing it to one here.

vulkaninfo/vulkaninfo.c
krOoze: Hence I commented the name of the variable is misleading.

The `vulkaninfo` output looks like this:

```
VK_LAYER_LUNARG_core_validation (LunarG Validation Layer) Vulkan version 1.1.101, layer version 1
        Layer Extensions        count = 1
                VK_EXT_debug_report                 : extension revision  6
        Devices         count = 1
                GPU id       : 0 (AMD Radeon R9 200 Series)
                Layer-Device Extensions count = 2
                        VK_EXT_debug_marker                 : extension revision  4
                        VK_EXT_validation_cache             : extension revision  1
```

I guess it would look weird if the layer name was repeated there instead of only "Layer-Device".

cube/cube.c
krOoze: No that does not help. I just point out the code rot. Were someone to change the variable to 2 (for any reason), then this code is wrong,
If there is not obvious immediate fix just disregard this in this PR.

vulkaninfo/vulkaninfo.c
krOoze: It does not matter. `VK_ERROR_INITIALIZATION_FAILED` or `VK_ERROR_MEMORY_MAP_FAILED` feels vaguely appropriate.
`return` too is ok I guess. Just have to be kept in mind were someone to refactor the error handling to support non-Vulkan errors....

jeremyk-lunarg:
Will merge after lunch if the author has no other changes they want to make.

PR #181, opened 2019-04-19
vulkaninfo: Check support for surface extensions before using them

polarina:
Not all ICDs support all the surface extensions that vulkaninfo
enumerates.

For example, on Linux with an ICD that only support VK_KHR_xcb_surface,
vulkaninfo crashes:

	Presentable Surfaces:
	=====================
	ERROR: VK_KHR_xlib_surface extension not enabled.  vkCreateXlibSurfaceKHR not executed!

	vulkaninfo/vulkaninfo.c:1427: failed with VK_ERROR_EXTENSION_NOT_PRESENT


jeremyk-lunarg:
Global CI should be fixed.  Please rebase and resubmit your change so that it can run through CI.

PR #186, opened 2019-05-01
vulkaninfo: Fix crash at Presentable Surface

jeremyk-lunarg:
Checks to make sure surface extensions were enabled
neglected to leave important struct variables outside
the if statements.

This resulted in these variables becoming uninitialized
and causing undefined behavior.

### DIFF
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -581,7 +581,7 @@ static const char *VkPresentModeString(VkPresentModeKHR mode) {
 
 static bool CheckExtensionEnabled(const char *extension_to_check, const char **extension_list, uint32_t extension_count) {
     for (uint32_t i = 0; i < extension_count; ++i) {
-        if (!strcmp(extension_to_check, extension_list[i])) {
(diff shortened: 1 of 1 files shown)


vulkaninfo/vulkaninfo.c
```diff
@@ -581,7 +581,7 @@ static const char *VkPresentModeString(VkPresentModeKHR mode) {
 
 static bool CheckExtensionEnabled(const char *extension_to_check, const char **extension_list, uint32_t extension_count) {
     for (uint32_t i = 0; i < extension_count; ++i) {
-        if (!strcmp(extension_to_check, extension_list[i])) {
+        if (!strncmp(extension_to_check, extension_list[i], VK_MAX_EXTENSION_NAME_SIZE)) {
```
lunarpapillo: I don't see why changing `strcmp()` to `strncmp()` is required here (and below)...?  It's not a bad idea in theory, but I don't think `CheckExtensionEnabled()` is ever called without a constant extension name, and I don't think this change will make a difference...?

Is it just a hygiene thing that got left around after making the real fix?

jeremyk-lunarg: It's mainly a hygiene thing.  I saw it and thought I would change it.
It's probably not necessary, but it's better safe than sorry.

PR #187, opened 2019-05-03
scripts: Update known good for 107 header update

shannon-lunarg:
Updated:
- `build-android/vulkan-headers_revision_android`
- `scripts/known_good.json`

These changes pass internal CI.


shannon-lunarg:
And now we have green lights across all GitHub checks.

PR #188, opened 2019-05-08
update_deps: make --clean-repo switch work in all cases

lunarpapillo:
--clean-repo right now will raise an exception if the repository
directory was not present.  This makes it work in all cases.

### DIFF
--- a/scripts/update_deps.py
+++ b/scripts/update_deps.py
@@ -343,7 +343,7 @@ class GoodRepo(object):
     def Checkout(self):
         print('Checking out {n} in {d}'.format(n=self.name, d=self.repo_dir))
         if self._args.do_clean_repo:
-            shutil.rmtree(self.repo_dir)
+            shutil.rmtree(self.repo_dir, ignore_errors=True)
         if not os.path.exists(os.path.join(self.repo_dir, '.git')):
             self.Clone()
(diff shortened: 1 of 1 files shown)


lunarpapillo:
So say the docs: 

> LGTM. I don't thing shutil.rmtree() throws other useful errors except to inform the user that the directory they are attempting to delete doesn't exist, but might be good to verify this before telling it to ignore all errors.

The docs say:

> If ignore_errors is true, errors resulting from failed removals will be ignored; if false or omitted, such errors are handled by calling a handler specified by onerror or, if that is omitted, they raise an exception.

Testing shows the only errors that are produced by `shutil.rmtree()` under normal system operation are:
```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/bb'
NotADirectoryError: [Errno 20] Not a directory: '/tmp/ee'
PermissionError: [Errno 1] Operation not permitted: '/tmp/ff'
```
All are ignored if `ignore_errors` is `True`.  The cases `NotADirectoryError` and `PermissionError` will both subsequently error out during `Clone()` or `Fetch()`; so the only change in behavior is if the directory doesn't exist.

jeremyk-lunarg:
Sounds good.  Thanks for checking.

PR #192, opened 2019-05-23
cmake: Use CMake 3.10.2 for CI

lunarpapillo:
These changes ensure that the Travis and AppVeyor
builds use a known version of CMake.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,17 @@ cache: ccache
 
 before_install:
   - set -e
+  - CMAKE_VERSION=3.10.2
+  - |
+    if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
+      CMAKE_URL="https://cmake.org/files/v${CMAKE_VERSION%.*}/cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz"
+      echo CMAKE_URL=${CMAKE_URL}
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -45,6 +45,17 @@ cache: ccache
...
+  - |
+    if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
+      CMAKE_URL="https://cmake.org/files/v${CMAKE_VERSION%.*}/cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz"
+      echo CMAKE_URL=${CMAKE_URL}
+      mkdir cmake-${CMAKE_VERSION} && travis_retry wget --no-check-certificate -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C cmake-${CMAKE_VERSION}
+      export PATH=${PWD}/cmake-${CMAKE_VERSION}/bin:${PATH}
+    else
+      brew install cmake || brew upgrade cmake
+    fi
+    cmake --version
```
jeremyk-lunarg: I don't see this appear in the Travis output.

lunarpapillo: Look for "Cmake version" in the raw log; you can see both the native version (3.12.4) and the newly installed version that will be used (3.10.2).  (I'm not sure how Travis "prettifies" its logs.)

jeremyk-lunarg: I see it now.  It was hidden under an expandable section.

Issue #122, opened 2019-01-14
vkcube: By which mechanism is the depth buffer synchronized


krOoze:
May I ask by which mechanism is the depth buffer synchronized in the `vkcube` demo?
This seems to be the source of many tutorials, but I suspect they just copy-paste this aspect without knowing why it works...

First use in frame seem to be the `loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR` (i.e. `VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT`), and last use the `storeOp = VK_ATTACHMENT_STORE_OP_DONT_CARE` (i.e. `VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT`).

The default `FRAME_LAG` is `2`, so it means there may purely theoretically be an overlap. I.e. fence wait is on an older frame, and not the immediately previous one.

There seem to be no subpass dependencies (except the implicitly defined ones).
The implicit renderpass external dependencies should do half the jorb. I.e. it  would make the depth buffer successfully "available from", but **not** "visible to" the ` loadOp` of the next frame.
Similarly, the implicit renderpass external dependencies of the next frame would make it "visible to", but not "available from" the previous frame. An execution dependency is missing between the two to make it work.

And there seem to be no pipeline barriers.

There is a semaphore signal on the batch, but only `vkQueuePresent` waits on it, which implies nothing about the state of the resource to the next frame. That is so because based on spec it does not cover subsequent submissions:

> Also, in the case of `vkQueueSubmit`, the second synchronization scope additionally includes all commands that occur later in submission order.

There is a semaphore wait on the batch. But the semaphore is from `vkAcquireNextImageKHR`, which does not imply much about the state of the draw of the previous frame. And the `pWaitDstStageMask` is `VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT` which does not match the depth buffer `loadOp` anyway.

I think the only way this would work is if there was some implicit relation of `vkQueuePresent` and `vkAcquireNextImageKHR`, that I cannot support from spec. (Or some other obscure mechanism from spec, which I missed, that is outside the realm of the conventional synchronization model).

Tony-LunarG:
Fixed with #200

PR #197, opened 2019-06-05
Cube: Add depth buffer per swapchain image

Tony-LunarG:
Addresses #122 by allocating a depth buffer for each swapchain image, removing the need for additional synchronization.  Also fixes a typo.

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -1828,14 +1828,14 @@ static void demo_prepare_descriptor_layout(struct demo *demo) {
     err = vkCreateDescriptorSetLayout(demo->device, &descriptor_layout, NULL, &demo->desc_layout);
     assert(!err);
 
-    const VkPipelineLayoutCreateInfo pPipelineLayoutCreateInfo = {
+    const VkPipelineLayoutCreateInfo pipelineLayoutCreateInfo = {
         .sType = VK_STRUCTURE_TYPE_PIPELINE_LAYOUT_CREATE_INFO,
         .pNext = NULL,
         .setLayoutCount = 1,
         .pSetLayouts = &demo->desc_layout,
     };
 
-    err = vkCreatePipelineLayout(demo->device, &pPipelineLayoutCreateInfo, NULL, &demo->pipeline_layout);
+    err = vkCreatePipelineLayout(demo->device, &pipelineLayoutCreateInfo, NULL, &demo->pipeline_layout);
     assert(!err);
 }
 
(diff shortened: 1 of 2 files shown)


Tobski:
@Tony-LunarG given that this cube demo is an oft-referenced example when developers are starting out, it would be good if it were following best practices - and this change is really not best practices. What's the motivation for adding multiple depth buffers instead of fixing the synchronization situation?

jzulauf-lunarg:
@tobski -- my thought was "toy solutions for toy problems." Given that for a zero stall solution (two depth buffers) this only requires an additional depth buffer to match with the 3 present buffers and leverage the existing synchronization code, it seemed reasonable.

If you have a suggested "best practice" fix (or approach), we'd certainly be open to it.

Tony-LunarG:
@Tobski would a barrier be more appropriate?  Is the barrier that I added to cube.c in the tony-cube-depth-barrier branch correct and sufficient?

cforfang:
> Given that for a zero stall solution (two depth buffers) this only requires an additional depth buffer to match with the 3 present buffers and leverage the existing synchronization code, it seemed reasonable.

I'm wondering what the thinking is for why you need two depth buffers to avoid stalls here? The GPU is only going to be rendering to one image at a time, so it's not like not having N depth-buffers is going to prevent much overlap. The only potential 'issue' I can think of is there will -- with 1 depth-buffer -- have to be a execution barrier between render-passes sharing it, to make sure the previous pass is done before the next starts, but _tripling_ memory consumption to avoid this kind of _super-tiny_ stall doesn't seem like good practice.

In fact, personally my main objection to this change is that it gives the impression you _want_ or _need_ to N-buffer in this case to avoid some kind of heavy performance hit, which is not the case. :)

Tobski:
Sorry I was off yesterday so couldn't respond, but basically what @cforfang said 😅

cforfang:
> @Tobski would a barrier be more appropriate? Is the barrier that I added to cube.c in the tony-cube-depth-barrier branch correct and sufficient?

Direct link: https://github.com/KhronosGroup/Vulkan-Tools/commit/8ed50eb17f012723b5dd42edd180e141f6c844ab

This looks pretty reasonable to me :) Possibly one would want to use `(VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT)` for the stages; @Tobski?

Tony-LunarG:
Closing in lieu of #200

Issue #179, opened 2019-04-15
Win10: vkcube with invalid arguments produces no output (graphics nor console)


mikew-lunarg:
When run from a console prompt, `vkcube -c 250`  does not open a graphics window, and no error text output; it just exits silently (The correct command is `vkcube --c 250`, which does run properly)  

Similarly, `vkcube foo` also exits silently with no graphics.

Other than the lack of help text for invalid argments, the reason this is an issue is the [GNU argument syntax](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) specifies "--" for long arguments, and '-' for single-character arguments.   Thus the `--c` argument _should_ be `-c`, and it is very easy to overlook the difference depending on the font.

Perhaps add `-c` as a synonym for `--c`.  It would be nice for help usage text when invalid arguments, but stdout on Windows adds complexity.

KarenGhavam-lunarG:
@mikew-lunarg Duplicate of #177 ?

davidlunarg:
No, I don't think it is a duplicate. This issue seems to say that if you give bad args to vkcube, it exits silently. #177 is a problem that when vkcube --c 250 runs, it runs so fast that nothing is seen.

mikew-lunarg:
David is correct.  Also, this one is Windows, #177 is Linux.  Different problem, different platform.

mikew-lunarg:
For comparison, on Linux a bad argument produces this text output on the terminal:
```
mikew@spork:build !15 1 $ ./vkcube -c 250
Usage:
  vkcube	[--use_staging] [--validate] [--validate-checks-disabled] [--break]
	[--c <framecount>] [--suppress_popups] [--incremental_present] [--display_timing]
	[--present_mode <present mode enum>]
	 <present_mode_enum>	VK_PRESENT_MODE_IMMEDIATE_KHR = 0
				VK_PRESENT_MODE_MAILBOX_KHR = 1
				VK_PRESENT_MODE_FIFO_KHR = 2
				VK_PRESENT_MODE_FIFO_RELAXED_KHR = 3
```

Tony-LunarG:
Fixed with #198

PR #198, opened 2019-06-06
cube: Add usage message on Windows

Tony-LunarG:
Fix for #179


jeremyk-lunarg:
cube.c looks good, but is there some way to us std::string objects in cubepp.cpp that would keep the code more C++-like?

Tony-LunarG:
stringstream version pushed

PR #200, opened 2019-06-12
cube: Use barrier to sync depth buffer

Tony-LunarG:
Because the contents of the depth buffer don't need to be preserved, a mere execution barrier should be sufficient.   @cforfang do the barrier stage masks look ok to you?

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -1,7 +1,7 @@
 /*
- * Copyright (c) 2015-2016 The Khronos Group Inc.
- * Copyright (c) 2015-2016 Valve Corporation
- * Copyright (c) 2015-2016 LunarG, Inc.
+ * Copyright (c) 2015-2019 The Khronos Group Inc.
+ * Copyright (c) 2015-2019 Valve Corporation
+ * Copyright (c) 2015-2019 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -1892,6 +1892,32 @@ static void demo_prepare_render_pass(struct demo *demo) {
         .preserveAttachmentCount = 0,
         .pPreserveAttachments = NULL,
     };
+
+    VkSubpassDependency attachmentDependencies[2] = {
+        [0] =
+            {
+                // Depth buffer is shared between swapchain images
+                .srcSubpass = VK_SUBPASS_EXTERNAL,
+                .dstSubpass = 0,
+                .srcStageMask = VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT,
+                .dstStageMask = VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT,
+                .srcAccessMask = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT,
+                .dstAccessMask = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT | VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT,
+                .dependencyFlags = 0,
+            },
+        [1] =
+            {
+                // Image Layout Transition
+                .srcSubpass = VK_SUBPASS_EXTERNAL,
+                .dstSubpass = 0,
+                .srcStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT,
+                .dstStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT,
+                .srcAccessMask = 0,
+                .dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT | VK_ACCESS_COLOR_ATTACHMENT_READ_BIT,
+                .dependencyFlags = 0,
+            },
+    };
+
     const VkRenderPassCreateInfo rp_info = {
         .sType = VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO,
         .pNext = NULL,
@@ -1900,8 +1926,8 @@ static void demo_prepare_render_pass(struct demo *demo) {
         .pAttachments = attachments,
         .subpassCount = 1,
         .pSubpasses = &subpass,
-        .dependencyCount = 0,
-        .pDependencies = NULL,
+        .dependencyCount = 2,
+        .pDependencies = attachmentDependencies,
     };
     VkResult U_ASSERT_ONLY err;
 
--- a/cube/cube.cpp
+++ b/cube/cube.cpp
@@ -1,7 +1,7 @@
 /*
- * Copyright (c) 2015-2016 The Khronos Group Inc.
- * Copyright (c) 2015-2016 Valve Corporation
- * Copyright (c) 2015-2016 LunarG, Inc.
+ * Copyright (c) 2015-2019 The Khronos Group Inc.
+ * Copyright (c) 2015-2019 Valve Corporation
+ * Copyright (c) 2015-2019 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -1975,13 +1975,33 @@ void Demo::prepare_render_pass() {
                              .setPreserveAttachmentCount(0)
                              .setPPreserveAttachments(nullptr);
 
+    vk::PipelineStageFlags stages = vk::PipelineStageFlagBits::eEarlyFragmentTests | vk::PipelineStageFlagBits::eLateFragmentTests;
+    vk::SubpassDependency const dependencies[2] = {
+        vk::SubpassDependency()  // Depth buffer is shared between swapchain images
+            .setSrcSubpass(VK_SUBPASS_EXTERNAL)
+            .setDstSubpass(0)
+            .setSrcStageMask(stages)
+            .setDstStageMask(stages)
+            .setSrcAccessMask(vk::AccessFlagBits::eDepthStencilAttachmentWrite)
+            .setDstAccessMask(vk::AccessFlagBits::eDepthStencilAttachmentRead | vk::AccessFlagBits::eDepthStencilAttachmentWrite)
+            .setDependencyFlags(vk::DependencyFlags()),
+        vk::SubpassDependency()  // Image layout transition
+            .setSrcSubpass(VK_SUBPASS_EXTERNAL)
+            .setDstSubpass(0)
+            .setSrcStageMask(vk::PipelineStageFlagBits::eColorAttachmentOutput)
+            .setDstStageMask(vk::PipelineStageFlagBits::eColorAttachmentOutput)
+            .setSrcAccessMask(vk::AccessFlagBits())
+            .setDstAccessMask(vk::AccessFlagBits::eColorAttachmentWrite | vk::AccessFlagBits::eColorAttachmentRead)
+            .setDependencyFlags(vk::DependencyFlags()),
+    };
+
     auto const rp_info = vk::RenderPassCreateInfo()
                              .setAttachmentCount(2)
                              .setPAttachments(attachments)
                              .setSubpassCount(1)
                              .setPSubpasses(&subpass)
-                             .setDependencyCount(0)
-                             .setPDependencies(nullptr);
+                             .setDependencyCount(2)
+                             .setPDependencies(dependencies);
 
     auto result = device.createRenderPass(&rp_info, nullptr, &render_pass);
     VERIFY(result == vk::Result::eSuccess);


mikew-lunarg:
Please bump copyright dates, they're way overdue.

cforfang:
I'm wondering if this needs a image barrier since it's a write-after-write dependency. Note regarding this in the spec says:
`write-after-write hazards need appropriate memory dependencies to be included
between them`. Otherwise it seems fine, though I'm not fully convinced about `LATE_FRAGMENT_TEST` -> `EARLY_FRAGMENT_TEST` vs `LATE/EARLY` -> `LATE/EARLY`. Personally I'd be conservative and have both on both sides for safety; it seems very unlikely to matter in practice though.

jzulauf-lunarg:
@Tobski -- given that we aren't interested in the last depth values (loadop clears it), am I correct that we don't an image memory barrier as the availability and visibility are irrelevant?

cforfang:
My thinking for why you might need one is that you have no idea how e.g. the clear will be done; it might be done by some other HW block which needs to have its caches invalidated to correctly read the (new) state of the image after the last render-pass completes (for example information about whether or not image has compression / HiZ applied, and this again informs how exactly to perform the clear). Note how the fact that a clear is done is different from a transition from undefined layout (which requires a image barrier anyway) which does in fact remove the need to consider the previous state.

jzulauf-lunarg:
@cforfang -- yeah, I'm looking for Tobias confirmation, but according to the spec it's the late fragment stage that does the storeop and early fragment that does that load op.  If the store is complete, and the loadop doesn't need the contents of the storeop... my understanding (if correct) is that the execution barrier should be enough.

cforfang:
Ah, re. the comment about UNDEFINED -- I missed that this patch does _not_ change `initialLayout = VK_IMAGE_LAYOUT_UNDEFINED` for the depth-buffer like the previous proposal did. In that case, there is definitely seems to be a problem because there's no synchronization around when this transition from `UNDEFINED` might take place (as there's no specified `VK_SUBPASS_EXTERNAL` `VkSubpassDependency`, the implicit one is used -- and this one has a barrier from `TOP -> *`, i.e. the transition from undefined can happen 'at any point').

With `initialLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL` like the old proposal had, there's no transition so this wasn't an issue.

jzulauf-lunarg:
@cforfang -- I think a layout transition from "UNDEFINED" to anything is simply "bookkeeping" with no access to the resource. As the previous contents are not required to be preserved, there's no need to rewrite the resource.

@tobski -- if there was some transition from something *other* than undefined *and* that we wanted to preserve ... what would be the appropriate stage mask be that would guard against the initial layout transition?

cforfang:
>  I think a layout transition from "UNDEFINED" to anything is simply "bookkeeping" with no access to the resource. As the previous contents are not required to be preserved, there's no need to rewrite the resource.

No need to rewrite the old content, but there is commonly a need to _initialize the new layout_ -- which can require actual memory writes, eg. initialization of image metadata related to compression or HiZ or whatever (the details are obviously GPU specific; the point is a transition is where the image is 'made ready' for whatever it is going to be used for next, eg. some fancy depth buffer format with HiZ for depth optimal layout, or a catchall-format all HW blocks support for general layout). This initialization is potentially actual 'work', and so if it isn't synchronized properly you risk the new layout initialization happening while the image is still being used.

The thing to remember is this type of initialization happens as part of the _transition_, not on first use. A barrier blocking the clear in early Z is only going to delay the first use, not the transition as this is governed by the external subpass dependency here.

HansKristian-Work:
The pipeline barrier looks wrong, the depth buffer sees a write-after-write hazard, and there is no memory barrier. From spec, there is a note clarifying write-after-write must be resolved:

```
Execution and memory dependencies are used to solve data hazards, i.e. to ensure
that read and write operations occur in a well-defined order. Write-after-read
hazards can be solved with just an execution dependency, but read-after-write and
write-after-write hazards need appropriate memory dependencies to be included
between them. If an application does not include dependencies to solve these
hazards, the results and execution orders of memory accesses are undefined.
```

Tobski:
It's great being able to just confirm on a bug what other people have said 😄 - you do need a memory barrier, as per comments from @cforfang and @HansKristian-Work.

With the memory barrier in place you'll need early/late on both sides, so that any writes to the depth attachment are covered - using a simple memory barrier is fine if the layout isn't changing and you already have a memory barrier in place.

jzulauf-lunarg:
@Tobski  @cforfang 

Not to throw a spanner into the process, but why aren't the implicit attachment barriers sufficient.  Given that with the following barrier and scope definitions... what am I missing? Depth is *only* being used as an attachment, the the two barrier seem to form a sufficient dependency chain.

(pardon the extended quote)

> If there is no subpass dependency from VK_SUBPASS_EXTERNAL to the first subpass that uses an attachment, then an implicit subpass dependency exists from VK_SUBPASS_EXTERNAL to the first subpass it is used in. The subpass dependency operates as if defined with the following parameters:

```
VkSubpassDependency implicitDependency = {
    .srcSubpass = VK_SUBPASS_EXTERNAL;
    .dstSubpass = firstSubpass; // First subpass attachment is used in
    .srcStageMask = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
    .dstStageMask = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
    .srcAccessMask = 0;
    .dstAccessMask = VK_ACCESS_INPUT_ATTACHMENT_READ_BIT |
                     VK_ACCESS_COLOR_ATTACHMENT_READ_BIT |
                     VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT |
                     VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT |
                     VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
    .dependencyFlags = 0;
};
```
> Similarly, if there is no subpass dependency from the last subpass that uses an attachment to VK_SUBPASS_EXTERNAL, then an implicit subpass dependency exists from the last subpass it is used in to VK_SUBPASS_EXTERNAL. The subpass dependency operates as if defined with the following parameters:
```
VkSubpassDependency implicitDependency = {
    .srcSubpass = lastSubpass; // Last subpass attachment is used in
    .dstSubpass = VK_SUBPASS_EXTERNAL;
    .srcStageMask = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
    .dstStageMask = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;
    .srcAccessMask = VK_ACCESS_INPUT_ATTACHMENT_READ_BIT |
                     VK_ACCESS_COLOR_ATTACHMENT_READ_BIT |
                     VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT |
                     VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT |
                     VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
    .dstAccessMask = 0;
    .dependencyFlags = 0;
};
```

A from earlier in the text:

> If srcSubpass is equal to VK_SUBPASS_EXTERNAL, the first synchronization scope includes commands that occur earlier in submission order than the vkCmdBeginRenderPass used to begin the render pass instance. ...  the first synchronization scope is limited to operations on the pipeline stages determined by the source stage mask specified by srcStageMask.

> If dstSubpass is equal to VK_SUBPASS_EXTERNAL, the second synchronization scope includes commands that occur later in submission order than the vkCmdEndRenderPass used to end the render pass instance. ... the second synchronization scope is limited to operations on the pipeline stages determined by the destination stage mask specified by dstStageMask.

> The first access scope is limited to access in the pipeline stages determined by the source stage mask specified by srcStageMask. It is also limited to access types in the source access mask specified by srcAccessMask.

> The second access scope is limited to access in the pipeline stages determined by the destination stage mask specified by dstStageMask. It is also limited to access types in the destination access mask specified by dstAccessMask.

cforfang:
The problem is basically `srcStageMask = TOP` here:
```
    .srcSubpass = VK_SUBPASS_EXTERNAL;
    .dstSubpass = firstSubpass; // First subpass attachment is used in
    .srcStageMask = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
```
The short version is, this dependency is what's blocking the automatic transition (which again blocks the actual subpass), so it _needs to be connected to the last use of the image_ for things to work right. In this case the last use of the image would've been in `VK_PIPELINE_STAGE_(LATE|EARLY)_FRAGMENT_TESTS_BIT`, so blocking on `TOP` isn't good enough.

The dependency from the last subpass in the previous use doesn't help either:
```    .srcSubpass = lastSubpass; // Last subpass attachment is used in
    .dstSubpass = VK_SUBPASS_EXTERNAL;
    .srcStageMask = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
    .dstStageMask = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;
```
This would only block the `BOTTOM` stage of upcoming work from starting, but the upcoming image-use happens in `VK_PIPELINE_STAGE_(LATE|EARLY)_FRAGMENT_TESTS_BIT` (or earlier if there's an automatic transition) -- which is _before_ `BOTTOM`.

Note finally how there's no real dependency chain here; you have `ALL -> BOTTOM` in frame N-1 and `TOP -> ALL` in frame N. `BOTTOM` and `TOP` doesn't connect. You could make it `ALL -> TOP` in frame N-1 and there _would_ be a dependency (`ALL -> TOP` -> `TOP -> ALL`), though it's be a very conservative one.

jzulauf-lunarg:
@cforfang  -- So it sounds like the barrier we need would be:

```
VkSubpassDependency depthAttachmentDependency = {
    .srcSubpass = VK_SUBPASS_EXTERNAL;
    .dstSubpass = 0;
    .srcStageMask = VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT; // Implies all earlier stages
    .dstStageMask = VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT; // Implies all later stages
    // Or do I need TOP to protect the layout transition?
    .srcAccessMask =  VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT; // Reads aren't an access hazard, given the execution barrier implied in the stage mask.
    .dstAccessMask = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT |
                     VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;  // need to protect any use 
    // (though with the loadOp=VK_ATTACHMENT_LOAD_OP_CLEAR, should there be any reads before writes?) Or is there a specific mask for the layout transition?
    .dependencyFlags = 0;
};
```

Does that look like a best practice barrier for the reuse of a depth buffer? @Tobski

cforfang:
That seems right to me with one exception; Tobias pointed out one does needs EARLY and LATE on both sides:

> With the memory barrier in place you'll need early/late on both sides of the execution barrier, so that any writes to the depth attachment are covered

I interpret this as 'both EARLY and LATE can write to the image, hence both stages must be part of the memory dependency' (and so explicitly listed in the stage-masks).

jzulauf-lunarg:
> I interpret this as 'both EARLY and LATE can write to the image, hence both stages must be part of the memory dependency' (and so explicitly listed in the stage-masks).

Because while synchronization scopes imply earlier (or later) for src (or dst) stages... the access scopes are *explicit* (i.e. only the referenced stages)

Correct?

cforfang:
Exactly :)

Tony-LunarG:
Updated with dependency instead of barrier

cforfang:
Looks good to me :+1:

jzulauf-lunarg:
@Tobski -- any chance you could add this example to your Synchronization Examples page?

cforfang:
Actually, one thing. There's also the issue of synchronizing the swap-chain image. It's basically the same deal; there needs to be a dependency chain to stop the transition from UNDEFINED from happening too early.

`pWaitDstStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT`, so the acquire-semaphore is signaled in that stage. So we need to make sure the transition doesn't happen before that stage.

So, should add `src/dstStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT` and `dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT`.

jzulauf-lunarg:
@cforfang -- the QueueSubmit wait semaphore has the stagemask you're recommending which is waiting for the present to be done using the attachment.  Are you saying we *additionally* need a sub-pass dependency, or that one would be preferable to the QueueSubmit wait semaphore?

cforfang:
> Are you saying we additionally need a sub-pass dependency

This, _additionally_. The semaphore only blocks `COLOR` but by default the transition can happen any time between `TOP` and  `COLOR`, i.e. it's unaffected by the semaphore. Then there's also the need to deal with the after-effects of the transition, as mentioned.

cforfang:
Just for reference, this is why the automatic transition can happen as early as it can:
> Automatic layout transitions away from initialLayout happens-after the availability operations for
all dependencies with a srcSubpass equal to VK_SUBPASS_EXTERNAL

In other words, it can happen at any time after the `VK_SUBPASS_EXTERNAL` dependency's `srcStageMask` -- which is `TOP` by default, and `EARLY_FRAGMENT_TESTS` in the current version. Both of these happens before the current `pWaitDstStageMask` (`COLOR`), hence the automatic transition is _not_ blocked on the semaphore. But it should be :)

jzulauf-lunarg:
Looking at the wiki -- can you explain why the `srcStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT` functions as it does *from* the spec?  I don't doubt it's correct.  I just can't see how it could be derived from the documentation.

>If srcSubpass is equal to VK_SUBPASS_EXTERNAL, the first synchronization scope includes commands that occur earlier in submission order than the vkCmdBeginRenderPass used to begin the render pass instance

> The first access scope is limited to access in the pipeline stages determined by the source stage mask specified by srcStageMask. It is also limited to access types in the source access mask specified by srcAccessMask.

In this text (and in all other places in the sync text) the srcStageMask applies to actions that happen *prior* -- indeed the scope is defined as "commands that occur earlier in submission order".  How then can that control the timing of a image layout transition (which doesn't happen at the color attachment stage) relative to a present (which *also* doesn't happen at the color attachment stage)?  

Where in the spec am I missing the definition of this behavior that says "if we give a srcStageMask for an image layout transition it limits the transition to wait for that same stage extant in the semaphores dstStageMask"  or defines whatever mechanism supports this behavior?

jzulauf-lunarg:
It's because Bs of the submit's semaphore wait intersects with the As of the Subpass dependency... isn't it?  The mental confusion is to extend the first sync scope definition of "commands that occur earlier in submission order" to include the semaphore wait of the queue submit.  Once one considers the sem wait as a prior command, the dependency chain comes into affect.  There is an execution sync on color attachment (from the queue submit "implicit command" of the semaphore wait), thus anything that subsequently has a source wait on that same stage is *also* delayed...

cube/cube.c
```diff
@@ -1900,8 +1912,8 @@ static void demo_prepare_render_pass(struct demo *demo) {
         .pAttachments = attachments,
         .subpassCount = 1,
         .pSubpasses = &subpass,
-        .dependencyCount = 0,
-        .pDependencies = NULL,
+        .dependencyCount = 1,
```
jzulauf-lunarg: Are we going to at the color attachment dependency in this PR?

cforfang:
Hmm, let me try to outline my thinking to see if it helps.

When setting `pWaitDstStageMask` we're creating an execution and memory dependency:
> When a batch is submitted to a queue via a queue submission, and it includes semaphores to be
waited on, it defines a memory dependency between prior semaphore signal operations and the
batch

The dependency has a synchronization and access scope. The first synchronization scope is the presentation engine being finished with the color image. We need to make sure any access to the image is part of the second synchronization scope (else we'd be accessing the image too early). 

> In the case of vkQueueSubmit, the second synchronization scope is limited to operations on the pipeline stages determined by the destination stage mask specified by the corresponding element of pWaitDstStageMask.

`pWaitDstStageMask` is here set to `VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT`, i.e. the second synchronization scope _only_ includes that (and logically later) stages.

So, to achieve what we want (no access to the swap-chain image before the semaphore signals) we need to somehow create a dependency between whatever the operations on the swap-chain image are and this `VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT`-only synchronization scope.

In this case, we're doing an automatic image transition on the image (because `initialLayout` != `layout` in the subpass), so _we need to make sure the transition is included in the synchronization scope_.

> Automatic layout transitions away from initialLayout happens-after the availability operations for
all dependencies with a srcSubpass equal to VK_SUBPASS_EXTERNAL

The spec tells us automatic transitions happens-after `VK_SUBPASS_EXTERNAL`-dependencies (that's _all we know about them_ and the only way to control _when_ they happen[1]). If we don't specify an explicit `VK_SUBPASS_EXTERNAL` dependency we get the implicit one with `srcSubpass` = `TOP`. In other words, with the implicit `VK_SUBPASS_EXTERNAL`-dependency the transition happens-after the `TOP` stage. In the current version of the code we specify a custom dependency, however, with `srcSubpass` = `EARLY_FRAMGENT_TEST`.

All we know about the automatic transition is it happens-after the stage we specify here. Neither stage is part of the semaphore signal's second synchronization scope (which is just `COLOR` remember). Hence, the transitions is not blocked on the semaphore signalling.

By including `COLOR` in `srcSubpass` we're making sure the transition will happen-after the `COLOR` stage. And as the `COLOR` stage _is part of the semaphore signal's sync scope_, we have a dependency chain: image use blocked on transition[2], transition blocked on `COLOR`, `COLOR` blocked on semaphore, semaphore blocked on presentation engine being finished with the image.

[1] This is why I don't like automatic image transitions; they feel very magical :) I prefer the more-explicit pipeline-barrier transitions.
[2] This goes back to the need for `dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT`, but I won't talk about that.

cforfang:
Ah, I missed the follow-up comment :D Well, shout if anything I said doesn't mesh with your current understanding :) Sounds like you got it.

Tony-LunarG:
Added second dependency for layout transition

cube/cube.c
```diff
@@ -1900,8 +1926,8 @@ static void demo_prepare_render_pass(struct demo *demo) {
         .pAttachments = attachments,
         .subpassCount = 1,
         .pSubpasses = &subpass,
-        .dependencyCount = 0,
-        .pDependencies = NULL,
+        .dependencyCount = 2,
+        .pDependencies = &attachmentDependencies,
```
jzulauf-lunarg: clang (correctly) complains about the & of the array  as opposed to the arrayname or the & of [0]

cforfang:
Hm, 1 combined dependency with `COLOR | EARLY | LATE` vs. `dependencyCount = 2` like you have here...

Seems fine, but I will point out at least RADV seems like it will overwrite the results of previously seen `VK_SUBPASS_EXTERNAL` dependencies if you have multiple such entries in `pDependencies` -- though that seems more like a driver bug than anything. (See https://github.com/intel/external-mesa/blob/master/src/amd/vulkan/radv_pass.c#L361 and note how it's _assigning_ rather than OR-ing.)

jzulauf-lunarg:
@cforfang -- I was hoping that by specify these separately, the depth dependency wouldn't be part of the dependency chain tied to the semaphore, which I think a combined dependency would require.

> a chain exists if the intersection of Bs in the first dependency and As in the second dependency is not an empty set. 

Certainly the implementation could validly OR these together, but I don't see that is a requirement.  But, then again, I've been surprised before.

cforfang:
> @cforfang -- I was hoping that by specifying these separately, the depth dependency wouldn't be part of the dependency chain tied to the semaphore, which I think a combined dependency would require.

Makes sense. FWIW, turns out the radv-thing is indeed a bug that's already fixed. LGTM.

PR #201, opened 2019-06-13
vulkaninfo: Fix HTML format for multi-GPU computer

jeremyk-lunarg:
Fixes an issue found where the device properties
and extensions would only be printed for the first GPU
on a system when printing in HTML format.


mikew-lunarg:
IIRC when we tested earlier. your first commit was 2 lines, the second was 1 line.
This has 3 lines; is this the 1-line fix applied 3 times?  Does that make sense?

jeremyk-lunarg:
Yes.  The problem was that some <detail> tags weren't getting closed properly.  There were 3 different areas this was happening. Not sure why it claimed there were only 2 changes or 1 change unless it somehow split up my incremental changes while you were looking at it.

PR #212, opened 2019-06-27
cube: Port Wayland impl from wl-shell to xdg-shell

mstoeckl:
This change ports vkcube in both C and C++ versions to use the stable
xdg-shell protocol for Wayland compositors. The original wl-shell
protocol has been deprecated and is being removed from major compositor
libraries; wlroots has already dropped support.

New cmake modules have been added to look for Wayland-Protocols
(containing XML descriptions of all common protocols) and
`wayland-scanner`, the tool used to convert the XML files to a usable C
interface.

The change also adds support for the xdg-decoration protocol, which for
some compositors is provided and needed to let them know that they
should draw a titlebar, borders with resize controls, and other standard
features.

Edit: as for the CLA, all changes in the pull request are placed in the public domain. I wrote the entire patch. If you still need the CLA to be signed despite that, it may take a few hours before I get around to reading it.

Also, it doesn't look like the CI actually checks code conditional on CUBE_WSI_SELECTION=WAYLAND, so it passing just means I haven't broken the XCB code by accident.

~Edit 2: And now that I try to sign the CLA, the website containing the form doesn't work. I'll try again later.~ Seems it did, good.

### DIFF
--- a/cmake/FindWaylandProtocols.cmake
+++ b/cmake/FindWaylandProtocols.cmake
@@ -1,0 +1,13 @@
+# Wayland protocols Defines:
+#
+# * WAYLAND_PROTOCOLS_FOUND    True if the wayland-protocols data path is found
+# * WAYLAND_PROTOCOLS_PATH     Path to the wayland-scanner executable
+#
+
+if(NOT WIN32)
+    find_package(PkgConfig)
+    pkg_check_modules(PKG_WAYLAND_PROTOCOLS QUIET wayland-protocols)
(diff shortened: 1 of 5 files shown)


jeremyk-lunarg:
@mstoeckl I'm sorry it has taken us so long to get to review this.
Please resolve the conflicts and we will review it once that is done.

Tony-LunarG:
Rebased, cherry picked and pushed to master

PR #213, opened 2019-06-27
vulkaninfo: Add '--list-formats' option

jeremyk-lunarg:
This commit makes the following changes:
  - Vulkaninfo no longer prints format properties by default.
  - Adds the `--list-formats` option to turn format listing back on.
  - Adds a description of `--list-formats` to `vulkaninfo.md`

### DIFF
--- a/vulkaninfo/vulkaninfo.c
+++ b/vulkaninfo/vulkaninfo.c
@@ -94,6 +94,7 @@ static bool html_output = false;
 static bool human_readable_output = true;
 static bool json_output = false;
 static uint32_t selected_gpu = 0;
+static bool list_formats = false;
 
 struct VkStructureHeader {
     VkStructureType sType;
@@ -2522,6 +2523,7 @@ void FormatPropsShortenedDump(const struct AppGpu *gpu) {
 }
 
 static void AppDevDump(const struct AppGpu *gpu, FILE *out) {
+    if (!list_formats && !json_output) return;
     if (html_output) {
         fprintf(out, "\t\t\t\t\t<details><summary>Format Properties</summary>\n");
     } else if (human_readable_output) {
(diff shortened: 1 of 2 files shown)


jeremyk-lunarg:
@MarkY-LunarG, that was the thing I wanted the most advice on.  My reasoning was that users who wanted the format info, once they see it gone, would Google what happened and find the feature.

I could also see it going the other way where someone gets frustrated with seeing formats all the time and looks up how to turn it off, but since formats have been on by default so far I don't think users will notice the change and thus not take advantage of it.

MarkY-LunarG:
I'm not disagreeing, I'm just asking if we've seen anything online about this.  I like it.  One other thing, should we make sure that the formats are still recorded if dumping out to JSON.  Or is that already done?

jeremyk-lunarg:
The formats should still be dumped if the output is set to JSON.

mikew-lunarg:
RE JSON: agree w @MarkY-LunarG 
Everything should be dumped when in JSON mode (well, the stuff for which JSON is already implemented), since there are simple tools for munging/filtering JSON as a postprocess.

karl-lunarg:
This is off-topic, but there is a missing HTML details closure for VkPhysicalDevicePCIBusInfoProperties.  The opening tag is at line 4488 in the current master branch.  Needs to have a closing tag after line 4504.  Might as well fix this too.

karl-lunarg:
I agree with the JSON direction of dumping all to the JSON file.  Probably should discuss the behavior for the HTML output as well.  

On one hand, if the Format Properties are always in the HTML file with the details tags, a user won't see the Format Properties unless they expand them in the browser, which is nice.  But if a browser doesn't support details or if Something Else is processing the HTML, then the user might want to generate an HTML file without the Format Properties and would, therefore, want them out by default and include them only when specifying the option.

This choice is slightly less obvious than the JSON one, but I'd propose always including the Format Properties in the HTML output, using the rationalization that the option really only affects "output to the screen" and that all "output to a file" behavior remains unchanged.  Like the JSON direction, there is no way to generate an HTML file without the Format Properties, but they can be removed by a script.

If you choose this direction, I think that the code needs to be changed to not return early if HTML output is selected.

charles-lunarg:
This functionality is in the refactor, and so this PR no longer needs to be merged in.

PR #216, opened 2019-07-02
build: Update cmake_minimum_required(VERSION 3.10.2)

mikew-lunarg:
Change-Id: Ie9e288bab742df9db8207db63a27661fb452fdfa

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -216,7 +216,7 @@ generate the native platform files.
     - [2017](https://www.visualstudio.com/vs/downloads/)
   - The Community Edition of each of the above versions is sufficient, as
     well as any more capable edition.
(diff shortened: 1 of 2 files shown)


jjulianoatnv:
Does this cause cmake to exit with a helpful error message when one's cmake version is older than the minimum?

I had an issue a couple months ago where I was unable to build (either Vulkan-Tools or Vulkan-Loader). It turned out my cmake version was too old. Updating cmake to a newer version fixed the issue.

mikew-lunarg:
@jjulianoatnv Yes it does.  Here's what I get with an older CMake:
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.10.2 or higher is required.  You are running version 3.4.1
```

PR #219, opened 2019-07-02
docs: Improve instructions for using update_deps.

jeremyk-lunarg:
[ci skip]

Same change as KhronosGroup/Vulkan-ValidationLayers#955

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -112,21 +112,40 @@ directories and place them in any location.
 
 ### Building Dependent Repositories with Known-Good Revisions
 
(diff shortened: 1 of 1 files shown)


jjulianoatnv:
This is a good improvement. It would have saved me some wasted effort when I built Vulkan-Tools recently on a new computer. I had made a mistake of specifying x64 when runninn update_deps.py, but then forgetting to specify x64 when doing the initial cmake command. Spelling out the steps as you are doing in this PR decreases probability that someone else will make the same mistake.

PR #226, opened 2019-07-23
Offline source code generation

### LINKED ISSUE #224: Bake codegen files into repo
Need code gen baking here as in VVL/Loader repos as mock_icd.h/cpp and vk_typemap_helper.h are generated from this repo.
This is a corollary issue to [965](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/965) in the VVL repo.
Code-gen files should be committed to repo, default build should use those files instead of running codegen, and CI should enforce that codegen files match committed files to prevent hand-editing of codegen.

mikes-lunarg:
Fixes #224 

This should all feel very familiar...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ project(Vulkan-Tools)
 
 # find_package(), include() and global project settings --------------------------------------------------------------------------
 
-find_package(PythonInterp 3 REQUIRED)
+find_package(PythonInterp 3 QUIET)
 
 # User-interface declarations ----------------------------------------------------------------------------------------------------
 # This section contains variables that affect development GUIs (e.g. CMake GUI and IDEs), such as option(), folders, and variables
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,6 +137,16 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
     endif()
 endif()
 
+# Optional codegen target
+if(PYTHONINTERP_FOUND)
+    add_custom_target(VulkanTools_generated_source
+                      COMMAND ${PYTHON_EXECUTABLE} ${PROJECT_SOURCE_DIR}/scripts/generate_source.py
+                              ${VulkanRegistry_DIR} --incremental
+                      )
+else()
+    message("WARNING: VulkanTools_generated_source target requires python 3")
+endif()
+
 if(APPLE)
     include(mac_common.cmake)
 endif()
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,8 +66,17 @@ That said, please ensure that the repository compiles and passes tests without e
 that to be accepted into the repository, the pull request must [pass all tests](#testing your changes) on all supported platforms
 -- the automatic Github Travis and AppVeyor continuous integration features will assist in enforcing this requirement.
 
+#### Generated Source Code
+
+The `icd/generated` directory contains source code that is created by several
+generator scripts in the `scripts` directory. All changes to these scripts _must_ be submitted with the
+corresponding generated output to keep the repository self-consistent. This requirement is enforced by both
+Travis CI and AppVeyor test configurations. Regenerate source files after modifying any of the generator
+scripts and before building and testing your changes. More details can be found in
+[BUILD.md](https://github.com/KhronosGroup/Vulkan-Tools/blob/master/BUILD.md#generated-source-code).
+
 #### **Testing Your Changes**
-* Run the repository components with the Vulkan Validation Layers before and after each if your commits to check for any regressions.
+* Run the repository components with the Vulkan Validation Layers before and after each of your commits to check for any regressions.
 
   (These instructions are for Linux)
 * In the `cube` directory, run:
--- a/scripts/generate_source.py
+++ b/scripts/generate_source.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+# Copyright (c) 2019 The Khronos Group Inc.
+# Copyright (c) 2019 Valve Corporation
+# Copyright (c) 2019 LunarG, Inc.
+# Copyright (c) 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author: Mike Schuchardt <mikes@lunarg.com>
+
+import argparse
+import filecmp
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+
+import common_codegen
+
+# files to exclude from --verify check
+verify_exclude = ['.clang-format']
+
+def main(argv):
+    parser = argparse.ArgumentParser(description='Generate source code for this repository')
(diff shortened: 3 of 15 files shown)


CMakeLists.txt
```diff
@@ -137,6 +137,16 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
...
 endif()
 
+# Optional codegen target
+if(PYTHONINTERP_FOUND)
+    add_custom_target(VulkanTools_generated_source
+                      COMMAND ${PYTHON_EXECUTABLE} ${PROJECT_SOURCE_DIR}/scripts/generate_source.py
+                              ${VulkanRegistry_DIR} --incremental
+                      )
+else()
+    message("WARNING: VulkanTools_generated_source target requires python 3")
```
lunarpapillo: I'm not sure `WARNING` is strong enough.  The `VulkanTools_generated_source` target should fail if Python3 is not available, yes?

CONTRIBUTING.md
```diff
@@ -66,6 +66,15 @@ That said, please ensure that the repository compiles and passes tests without e
...
+
+The `icd/generated` directory contains source code that is created by several
+generator scripts in the `scripts` directory. All changes to these scripts _must_ be submitted with the
+corresponding generated output to keep the repository self-consistent. This requirement is enforced by both
+Travis CI and AppVeyor test configurations. Regenerate source files after modifying any of the generator
+scripts and before building and testing your changes. More details can be found in
+[BUILD.md](https://github.com/KhronosGroup/Vulkan-Tools/blob/master/BUILD.md#generated-source-code).
+
 #### **Testing Your Changes**
 * Run the repository components with the Vulkan Validation Layers before and after each if your commits to check for any regressions.
```
lunarpapillo: Not you, but "each if your commits" => "each of your commits"

This can be deferred until the next change of this file, if we add an issue to track it.

scripts/generate_source.py
```diff
@@ -0,0 +1,111 @@
...
+    else:
+        # generate directly in the repo
+        gen_dir = repo_dir
+
+    # run each code generator
+    for cmd in gen_cmds:
+        print(' '.join(cmd))
+        try:
+            subprocess.check_call([sys.executable] + cmd,
+                                  # ignore generator output, vk_validation_stats.py is especially noisy
```
tobine: Irrelevant comment on vk_validation_stats for this repo

scripts/generate_source.py
```diff
@@ -0,0 +1,111 @@
...
+                                  stdout=subprocess.DEVNULL,
+                                  cwd=gen_dir)
+        except Exception as e:
+            print('ERROR:', str(e))
```
lunarpapillo: If an exception does occur, what's the expected developer response?

I think the developer will have to examine the script, figure out by trial and error which `check_call()` execution failed, look at the output, and then figure out what went wrong and fix it.

It would be easier if the `check_call()` saved its stdout and stderr, and printed the command line, working directory, any interesting environment variables, return code, and stdout and stderr in the case of error.  Then the developer wouldn't have to dissect the script and do experiments to figure out what happened.

scripts/kvt_genvk.py
```diff
@@ -26,15 +26,14 @@ def startTimer(timeit):
...
-    endTime = time.clock()
     if timeit:
         endTime = time.process_time()
         write(msg, endTime - startTime, file=sys.stderr)
         startTime = None
 
 # Turn a list of strings into a regexp string matching exactly those strings
 def makeREstring(list, default = None):
-    if len(list) > 0 or default == None:
+    if len(list) > 0 or default is None:
```
lunarpapillo: mild suggestion: this is slightly more Pythonic:
```
if list or default is None:
```

scripts/generate_source.py
```diff
@@ -0,0 +1,111 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author: Mike Schuchardt <mikes@lunarg.com>
+
+import argparse
+import common_codegen
```
lunarpapillo: mild: local imports should be in a separate section, e.g.:
```
import argparse
import filecmp
...
import tempfile

import common_codegen

```

scripts/generate_source.py
mikes-lunarg: I left it in the loader script too, my rationale was to minimize diffs across the repos.  We might even factor the command list into another file so the script can be identical everywhere.

scripts/kvt_genvk.py
mikes-lunarg: This file is forked from the upstream Vulkan-Docs repo https://github.com/KhronosGroup/Vulkan-Docs/blob/7b81afeb9f76ed8b1ca6c2ed9c93d94eee08bfed/scripts/genvk.py#L53

CMakeLists.txt
mikes-lunarg: You just wouldn't get a `VulkanTools_generated_source` target if you don't have Python3.  This is similar to the way LunarG/VulkanTools handles the optional Qt5 dependency for the vkconfig target: https://github.com/LunarG/VulkanTools/blob/8792090d531807522c808663b10043da4a2f502c/vkconfig/CMakeLists.txt#L12

scripts/generate_source.py
mikes-lunarg: I don't think the developer will have to do very much archaeology to find the error location.  The command is printed to stdout before executing, and stderr is unperturbed.  Here's what happens if I put a whoops in mock_icd_generator.py

```
$ make VulkanTools_generated_source   
/home/mike/Projects/Vulkan-Tools/scripts/kvt_genvk.py -registry /home/mike/Projects/Vulkan-Headers/build/install/share/vulkan/registry/vk.xml -quiet vk_typemap_helper.h
/home/mike/Projects/Vulkan-Tools/scripts/kvt_genvk.py -registry /home/mike/Projects/Vulkan-Headers/build/install/share/vulkan/registry/vk.xml -quiet mock_icd.h
Traceback (most recent call last):
  File "/home/mike/Projects/Vulkan-Tools/scripts/kvt_genvk.py", line 356, in <module>
    genTarget(args)
  File "/home/mike/Projects/Vulkan-Tools/scripts/kvt_genvk.py", line 225, in genTarget
    reg.apiGen(options)
  File "/home/mike/Projects/Vulkan-Headers/build/install/share/vulkan/registry/reg.py", line 1061, in apiGen
    self.gen.beginFile(self.genOpts)
  File "/home/mike/Projects/Vulkan-Tools/scripts/mock_icd_generator.py", line 1078, in beginFile
    raise Exception('whoops')
Exception: whoops
ERROR: Command '['/usr/bin/python3', '/home/mike/Projects/Vulkan-Tools/scripts/kvt_genvk.py', '-registry', '/home/mike/Projects/Vulkan-Headers/build/install/share/vulkan/registry/vk.xml', '-quiet', 'mock_icd.h']' returned non-zero exit status 1.
make[3]: *** [CMakeFiles/VulkanTools_generated_source.dir/build.make:57: CMakeFiles/VulkanTools_generated_source] Error 1
make[2]: *** [CMakeFiles/Makefile2:73: CMakeFiles/VulkanTools_generated_source.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:80: CMakeFiles/VulkanTools_generated_source.dir/rule] Error 2
make: *** [Makefile:164: VulkanTools_generated_source] Error 2
```

CONTRIBUTING.md
mikes-lunarg: Will fix while I'm here

scripts/generate_source.py
mikes-lunarg: Good code style suggestion

CONTRIBUTING.md
mikes-lunarg: Also fixed in VVL and VulkanTools

PR #246, opened 2019-08-21
build: Update known good for 1.1.120 vulkan header

shannon-lunarg:
Changes:
- Updated known good files
- Generated new files using `generate_source.py`

Updated:
- `build-android/vulkan-headers_revision_android
- `icd/generated/mock_icd.h`
- `scripts/known_good.json`

Change-Id: Ia83e6bc8acd0c54d23e97e9c645d0062014f4840


shannon-lunarg:
This passes internal and external CI

scripts/known_good.json
```diff
@@ -17,7 +17,7 @@
       "sub_dir" : "Vulkan-Headers",
       "build_dir" : "Vulkan-Headers/build",
       "install_dir" : "Vulkan-Headers/build/install",
-      "commit" : "v1.1.119"
+      "commit" : "9eab96f40634c81ed53d54d2c9df126a445dfd56"
```
jeremyk-lunarg: Why is this a commit ID and not a tag?

shannon-lunarg: Because the tag points to a commit that produces build errors when generating source scripts in all the repos and the SHA indicates the commit that fixes it.

jeremyk-lunarg: Sounds good!

mikew-lunarg: Ideally there should be a new tag, indicating the best commit to grab for the headers (and their related resources; I realize the headers themselves weren't changed)

shannon-lunarg: Indicating the best commits to use for building a repo is the purpose of the known-good file.

mikew-lunarg: Yes, and the other person who experienced the error.
Tags are used to indicate "immutable commits of interest".
I'd assert that every reference in known-good files, by definition, should be a tag.

PR #248, opened 2019-08-22
cube: Add support for VK_EXT_headless_surface

rosen-zhelev:
Adds HEADLESS as an option for CUBE_WSI_SELECTION
This builds cube using the VK_EXT_headless_surface
extension.

Change-Id: I3ec0d323fce996f15f3f85682a68758b1a213408

### DIFF
--- a/cube/CMakeLists.txt
+++ b/cube/CMakeLists.txt
@@ -103,6 +103,8 @@ elseif(UNIX AND NOT APPLE) # i.e. Linux
         add_definitions(-DVK_USE_PLATFORM_WAYLAND_KHR)
     elseif(CUBE_WSI_SELECTION STREQUAL "DISPLAY")
         add_definitions(-DVK_USE_PLATFORM_DISPLAY_KHR)
+    elseif(CUBE_WSI_SELECTION STREQUAL "HEADLESS")
+        add_definitions(-DVK_USE_PLATFORM_HEADLESS_EXT)
     else()
         message(FATAL_ERROR "Unrecognized value for CUBE_WSI_SELECTION: ${CUBE_WSI_SELECTION}")
     endif()
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -2921,6 +2921,17 @@ static void demo_run_display(struct demo *demo) {
         }
     }
 }
+#elif defined(VK_USE_PLATFORM_HEADLESS_EXT)
+static void demo_run_headless(struct demo *demo) {
+    while (!demo->quit) {
+        demo_draw(demo);
+        demo->curFrame++;
+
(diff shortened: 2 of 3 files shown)


jeremyk-lunarg:
The spec says with regards to VK_EXT_headless_surface:

> The VK_EXT_headless_surface extension is an instance extension. It provides a mechanism to create VkSurfaceKHR objects independently of any window system or display device. **The presentation operation for a swapchain created from a headless surface is by default a no-op, resulting in no externally-visible result**.

[Source](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VK_EXT_headless_surface)

Keeping in mind this quote and the fact that cube's entire point is to display an animated cube to the user, what is the reasoning for adding an option to cube that explicitly prevents it from displaying its images?

rosen-zhelev:
The main reason for this request is to have a sample application that can be used to verify support for the headless surface extension. We would like to use a sample application in testing a Vulkan layer which can implement WSI support: https://gitlab.freedesktop.org/mesa/vulkan-wsi-layer

I do appreciate that most of cube's usefulness is lost when using this WSI extension, but it seemed to be the simplest sample app for which to add headless WSI support. In addition a frame capture layer can be used to obtain the output of any application that is using the headless extension. Please correct me if there is a more appropriate Vulkan sample app for this.

jeremyk-lunarg:
The Vulkan Info application in this same repo should be able to verify whether an extension is supported or not.
This should include `VK_EXT_headless_surface`.

rosen-zhelev:
Vulkan Info can check if the headless surface extension is reported as supported, but would not use it to present any Vulkan rendered images. The advantage of having a sample application that uses headless surface to present its rendering is to demonstrate how Vulkan applications can work on a headless system. It can serve as an example of working with Vulkan on headless platforms.

If you think there is no value of using cube for this purpose then feel free to reject this request.

karl-lunarg:
Yes, it would probably be better to demonstrate the headless surface extension with its own dedicated sample.  A great deal of the code in `cube` needs to deal with the various windowing systems and all the details of updating the window.  I think that the headless additions would easily get lost among all of this.

Perhaps a more valuable approach would be to make a simple demo that draws a simple scene like a color interpolated triangle and then dump the contents of the target into a PPM file or something. That way, the user could actually "see" the results of the headless rendering.  The headless aspects would also be more "front and center".

As far as distribution goes, you could try a PR to the [LunarG Samples repo](https://github.com/LunarG/VulkanSamples).  There's also a [discussion](https://github.com/KhronosGroup/Vulkan-Ecosystem/issues/28) going on within Khronos that suggests they might revive a samples repo.

PR #255, opened 2019-09-05
build: Update glslang version in known-good.json

jeremyk-lunarg:


mikew-lunarg:
Don't we get a tag into the glslang repo?

jeremyk-lunarg:
We do not.

Issue #252, opened 2019-08-29
cube: not explicitly handling vkAcquireNextImageKHR returning VK_ERROR_SURFACE_LOST_KHR


danginsburg:
I have been working on debugging a problem on ARM Mali/Android 9 with my swapchain handling code in my engine.  I had based some of my code off of the cube example, and had been somewhat led astray by this code here:

https://github.com/KhronosGroup/Vulkan-Tools/blob/5e453384b73751d471467b34bb3e4e19e1073b96/cube/cube.c#L1009

I'm not sure if it happens with the cube sample on ARM Mali, but at least in my own code there are cases where VK_ERROR_SURFACE_LOST_KHR is returned by vkAcquireNextImageKHR and handling that case is required.

Now, granted - the cube code will work, although it will assert.  I think it would be better though if it called out that as a case since this serves as example code on how to handle the swapchain (especially with it being referenced as the canonical example of how to use VK_GOOGLE_display_timing).

Tony-LunarG:
Fixed via #259

PR #259, opened 2019-09-11
cube: Handle SURFACE_LOST return code

Tony-LunarG:
Addresses #252

### DIFF
--- a/cube/cube.cpp
+++ b/cube/cube.cpp
@@ -242,6 +242,7 @@ struct Demo {
     void prepare_textures();
 
     void resize();
+    void create_surface();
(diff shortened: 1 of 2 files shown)


cube/cube.cpp
```diff
@@ -731,6 +732,10 @@ void Demo::draw() {
             // swapchain is not as optimal as it could be, but the platform's
             // presentation engine will still present the image correctly.
             break;
+        } else if (true) {  //(result == vk::Result::eErrorSurfaceLostKHR) {
```
jeremy-lunarg: else if true?

Tony-LunarG: My bad - test code.  Fixed

PR #265, opened 2019-10-05
cube: Try to bind to wl_compositor v4

GeorgesStavracas:
VK_KHR_incremental_present requires the client to
bind to wl_compositor v4. This is because internally
some Vulkan Wayland drivers convert VkPresentRegionKHR
to wl_surface.damage_buffer() [1].

If the Wayland compositor advertises wl_compositor v4,
we can bind to it and make incremental present actually
work.

[1] https://gitlab.freedesktop.org/mesa/mesa/blob/master/src/vulkan/wsi/wsi_common_wayland.c#L862

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -3636,12 +3636,16 @@ static const struct wl_seat_listener seat_listener = {
     seat_handle_capabilities,
 };
 
-static void registry_handle_global(void *data, struct wl_registry *registry, uint32_t id, const char *interface,
-                                   uint32_t version UNUSED) {
+static void registry_handle_global(void *data, struct wl_registry *registry, uint32_t id, const char *interface, uint32_t version) {
     struct demo *demo = data;
     // pickup wayland objects when they appear
     if (strcmp(interface, "wl_compositor") == 0) {
-        demo->compositor = wl_registry_bind(registry, id, &wl_compositor_interface, 1);
+        uint32_t minVersion = version < 4 ? version : 4;
+        demo->compositor = wl_registry_bind(registry, id, &wl_compositor_interface, minVersion);
+        if (demo->VK_KHR_incremental_present_enabled && minVersion < 4) {
+            fprintf(stderr, "Wayland compositor doesn't support VK_KHR_incremental_present, disabling.\n");
+            demo->VK_KHR_incremental_present_enabled = false;
+        }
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
I have adapted the code style as per Travis CI's complaint. This issue was found while working on [GNOME/gtk!1116](https://gitlab.gnome.org/GNOME/gtk/merge_requests/1116). Let me know if there's anything else that is needed for merging.

Tony-LunarG:
So are you saying that there is a case where the driver supports incremental_present, but the compositor does not?  Is version 4 the right version for all Wayland compositors or just one specific compositor?  Is this in any documentation that you  could point me to?  Shouldn't the driver be the place where things like this are handled?  Please forgive my lack of Wayland knowledge.

GeorgesStavracas:
> So are you saying that there is a case where the driver supports incremental_present, but the compositor does not?

Yes; Wayland protocols are versioned, and each interface must have a version. When new requests, properties or signals are added to an interface, the version number is incremented. Wayland compositors must implement all properties, requests and signals that a certain interface version has in order to advertise support for that interface version.

Wayland clients and compositors then negotiate the interface version; the compositor tells the client which interface version it supports, and the client may bind to that version, or a lower version.

In this case, the interface we're looking at is `wl_compositor` [1] and `wl_surface` [2]. These two interfaces always have their versions in sync.

The incremental present regions are converted to a call to `wl_surface.damage_buffer()` [3], which is a v4 method. That means that only compositors that implement `wl_surface` (and `wl_compositor`) are able to handle it. Also, the client needs to **ask** the compositor to use v4 of `wl_surface`.

> Is version 4 the right version for all Wayland compositors or just one specific compositor?

Yes :)

The version number of any interface is given by the corresponding Wayland protocol (the XML file). Wayland compositors only tell whether they support that version or not.

> Shouldn't the driver be the place where things like this are handled?

The driver does need to check if the correct version of the protocol is supported, and ignore incremental present if it's not. This is the case for Intel Vulkan driver.

---

[1] https://gitlab.freedesktop.org/wayland/wayland/blob/master/protocol/wayland.xml#L190
[2] https://gitlab.freedesktop.org/wayland/wayland/blob/master/protocol/wayland.xml#L1323
[3] https://gitlab.freedesktop.org/wayland/wayland/blob/master/protocol/wayland.xml#L1677

PR #266, opened 2019-10-07
vulkaninfo: Fix closing immediately on windows

charles-lunarg:
On windows the console window will immediately close after running
vulkaninfo, this fixes it.


jeremyk-lunarg:
It would probably be better to put the second commit in a new PR, as it fixes a different problem than the one this PR is supposed to fix.

PR #267, opened 2019-10-08
vulkaninfo: Handle device group extension not available

charles-lunarg:
Previously if the device didn't support VK_KHR_device_group vulkaninfo
would not catch it and subsequently crash. This commit adds error
checking and bool flag to GetGroupCapabilities for callers to know
if it was successful.

files modified:
 - vulkaninfo/vulkaninfo.cpp
 - vulkaninfo/vulkaninfo.h

Change-Id: Ie8fe9cc46ebe0fa6cae9985d89628d60ddf58cf2

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -204,38 +204,41 @@ void DumpGroups(Printer &p, AppInstance &inst) {
             p.AddNewline();
 
             p.ObjectStart("Device Group Present Capabilities (Group " + std::to_string(group_id) + ")");
-
             auto group_capabilities = GetGroupCapabilities(inst, group);
-            for (uint32_t i = 0; i < group.physicalDeviceCount; i++) {
-                std::string device_out;
-                if (p.Type() == OutputType::text) {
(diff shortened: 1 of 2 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -204,38 +204,41 @@ void DumpGroups(Printer &p, AppInstance &inst) {
...
+                DumpVkDeviceGroupPresentModeFlagsKHR(p, "Present modes", group_capabilities.second.modes);
             }
-            DumpVkDeviceGroupPresentModeFlagsKHR(p, "Present modes", group_capabilities.modes);
             p.ObjectEnd();
             p.AddNewline();
-            group_id++;
         }
-        p.ObjectEnd();
-        p.AddNewline();
+        group_id++;
```
jeremyk-lunarg: This is no longer hit between info dumping for groups.
The consequence of this is that the group number in the output is never incremented.

charles-lunarg: Fixed, thanks for catching it.

PR #268, opened 2019-10-11
vulkaninfo: updated json documentation

charles-lunarg:
Currently, the json option doesn't specify that it only outputs for one
gpu and that it follows a schema. This adddresses that. In the future
there might be a PR which lets json output multigpu and all the same 
info as text/html, but for now its important to document the current state.

file changed
 - vulkaninfo/vulkaninfo.cpp

Change-Id: I3af8d559b2464d7ce00ce21e58a1c3054c0740b4

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -619,8 +619,8 @@ void print_usage(const char *argv0) {
     std::cout << "--html                Produce an html version of vulkaninfo output, saved as\n";
     std::cout << "                      \"vulkaninfo.html\" in the directory in which the command is\n";
     std::cout << "                      run.\n";
-    std::cout << "-j, --json            Produce a json version of vulkaninfo output to standard\n";
-    std::cout << "                      output.\n";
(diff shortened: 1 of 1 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -619,8 +619,8 @@ void print_usage(const char *argv0) {
     std::cout << "--html                Produce an html version of vulkaninfo output, saved as\n";
     std::cout << "                      \"vulkaninfo.html\" in the directory in which the command is\n";
     std::cout << "                      run.\n";
-    std::cout << "-j, --json            Produce a json version of vulkaninfo output to standard\n";
-    std::cout << "                      output.\n";
+    std::cout << "-j, --json            Produce a json version of vulkaninfo to standard output of the\n";
+    std::cout << "                      first gpu in the system following the dev-sim schema.\n";
```
mikew-lunarg: Suggest "conforming to the DevSim schema"

charles-lunarg: Thanks, that is much clearer.

PR #277, opened 2019-10-28
build: Update glslang version in known-good.json

jeremyk-lunarg:


jeremyk-lunarg:
Talked with @shannon-lunarg offline.
She will be including an update to the glslang version in her known-good update [here](https://github.com/KhronosGroup/Vulkan-Tools/pull/278).

PR #288, opened 2019-11-01
vulkaninfo: make where output goes uniform

charles-lunarg:
Currently, text and json output are put to standard output while
html is put to vulkaninfo.html. This commit makes it uniform so that
unless specifed, all output types are put to standard out. By using
the `--dest=<filename>` command line argument a user can manually
specify the file name of where to put the output.
In this implementation, the file is appended with the extension
appropriate for the output, so .txt for the default text, .html for
html, and .json for json.

This is a breaking change in that --html will not create a new file,
but this brings the application back in line with user expectations.

files modified
 - vulkaninfo/vulkaninfo.cpp
 - vulkaninfo/vulkaninfo.md

Change-Id: I97004a1743e13f938b4b0210ad8373abc5e57dd3

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -659,13 +659,11 @@ void print_usage(const char *argv0) {
     std::cout << "USAGE: " << argv0 << " [options]\n\n";
     std::cout << "OPTIONS:\n";
     std::cout << "-h, --help            Print this help.\n";
-    std::cout << "--html                Produce an html version of vulkaninfo output, saved as\n";
-    std::cout << "                      \"vulkaninfo.html\" in the directory in which the command is\n";
-    std::cout << "                      run.\n";
-    std::cout << "-j, --json            Produce a json version of vulkaninfo to standard output of the\n";
-    std::cout << "                      first gpu in the system conforming to the DevSim schema.\n";
-    std::cout << "--json=<gpu-number>   For a multi-gpu system, a single gpu can be targetted by\n";
-    std::cout << "                      specifying the gpu-number associated with the gpu of \n";
+    std::cout << "--html                Produces an html version of vulkaninfo to standard output.\n";
+    std::cout << "-j, --json            Produces a json version of vulkaninfo to standard output of\n";
+    std::cout << "                      the first gpu in the system conforming to the DevSim schema.\n";
+    std::cout << "--json=<gpu-number>   For a multi-gpu system, a single gpu can be targeted by\n";
+    std::cout << "                      specifying the gpu-number associated with the gpu of\n";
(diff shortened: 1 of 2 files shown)


mikew-lunarg:
+1 for all output to stdout.

I'm not convinced of the value of the new --dest= feature.  Is there anything it can do that can't already be accomplished with shell redirection?  It seems a needless and non-standard complication.

charles-lunarg:
I was under the impression that directly writing to a file is more useful for other applications that use vulkaninfo. While its often possible to redirect standard output inside another application, its usually simpler to execute vulkaninfo, then read the contents of the file it generated.

I'm not opposed to removing "--dest=", but would like to be sure that anyone one depending on a file being generated can switch to shell redirection.

mikew-lunarg:
Not disputing the value of writing to a file.

But that can already be done trivially w shell redirection; e.g. `vulkaninfo --help > myfile.html`

If there are usecases where redirection is not possible, then --dest= provides value.

mikew-lunarg:
Also, if redirection by the tool is necessary, the common switch for that is `-o` and `--output=`

lunarpapillo:
> I was under the impression that directly writing to a file is more useful for other applications that use vulkaninfo.

Looking at the alternatives for programs that wish to capture and process `vulkaninfo` output...
- Script programs (bash, cmd, Powershell, Python, perl)  are designed with ways to capture and process output.  There's no difficulty in either redirecting to a file, or capturing output directly without going to a file (the latter is generally considered more sophisticated).  So it doesn't matter there.
- Compiled programs can `fork()` and `exec()` to invoke a subprocess.  The proper way to redirect `stdout`, if necessary, is to close `stdout` and reopen it as desired in the subprocess.  Otherwise, the program will have to invoke a shell directly, and construct a command line that includes appropriate shell redirection operators.  The former is relatively easy and clean; the latter (which you may be thinking of) is a bear.

Are there any other use cases you're thinking of that I haven't considered?

Adding `--dest` does indeed simplify the "wrong" way to capture `stdout`; but I'm not sure the limited value of such is worth additional code complexity.

charles-lunarg:
I have updated to remove the file output entirely, relying on the user to pipe the output into a file if that is desired.

charles-lunarg:
Since this PR breaks vkconfig, I am waiting on an appropriate fix in vkconfig before merging.

KarenGhavam-lunarG:
Is there a corresponding VulkanTools issue for this required change for vkconfig?

charles-lunarg:
There is now.

lunarpapillo:
Sorry about that last review... apparently it was pending and I'd never submitted it?

charles-lunarg:
Closing branch due to changing requirements in vkconfig.

PR #292, opened 2019-11-14
vulkaninfo: return properly on xcb onnection errors

chivakker:
Calling xcb_disconnect on the connection to release memory and
return.

Otherwise on a headless display with no X11 backend the nullptr
for the connection will cause segmentation fault on following
xcb calls when xcb_connection_has_error returns a code.

examples ran on headless displays

DISPLAY=:0 vulkaninfo returns XCB_CONN_ERROR
DISPLAY=foobar vulkaninfo returns XCB_CONN_CLOSED_PARSE_ERR

Signed-off-by: Daniel Charles <daniel.charles@intel.com>

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -503,7 +503,9 @@ static void AppCreateXcbWindow(AppInstance &inst) {
     if (conn_error) {
         fprintf(stderr, "XCB failed to connect to the X server due to error:%d.\n", conn_error);
         fflush(stderr);
+        xcb_disconnect(inst.xcb_connection);
         inst.xcb_connection = nullptr;
+        return;
     }
 
     setup = xcb_get_setup(inst.xcb_connection);


vulkaninfo/vulkaninfo.h
```diff
@@ -503,7 +503,8 @@ static void AppCreateXcbWindow(AppInstance &inst) {
     if (conn_error) {
         fprintf(stderr, "XCB failed to connect to the X server due to error:%d.\n", conn_error);
         fflush(stderr);
-        inst.xcb_connection = nullptr;
```
charles-lunarg: I see that `xcb_disconnect(inst.xcb_connection);` needs to be called for freeing memory, but I fail to see why the nulling out of the inst.xcb_connection should be removed. It serves as a guard for later code to insure nothing uses the invalid xcb_connection.

chivakker: On line 500 above it says API guarantees non-null xcb_connection, shouldn't that be taken care of elsewhere?

charles-lunarg: The comment refers to xcb_connect() never returning a null pointer, so unless we also kept the conn_error around, the only way to know if it is valid is to null it out when it fails. Code later on, namely AppCreateXcbSurface, need to know that and thus checks if xcb_connection is null.
xcb_disconnect() doesn't null out the pointer (since its pass-by-value), so we need to do it ourselves.

chivakker: ok, I can restore it I would like to confirm that nothing else will assume xcb_connection is not null like xcb_get_setup did.

charles-lunarg: That should be good.

charles-lunarg: And yes, AppCreateXcbSurface is the only function that is affected (that mattes) where the nullptr is needed for correct behavior.

chivakker: I pushed an updated version, let me know if something else has to be addressed.

PR #293, opened 2019-11-18
vulkaninfo: check that extensions aren't disabled

charles-lunarg:
Previous versions of vulkainfo's autogen code didn't check if enums and
bitmasks from extensions were disabled, possibly letting them leak into
the generated code. This commit checks and skips any fields which aren't
supported.

Also discovered that the `for ext in extensions:` block in VulkanEnum and VulkanBitmask was never being used, and removed it as it didn't affect the generated output in the slightest. 

File changed:
	scripts/vulkaninfo_generator.py

Change-Id: I0e6fd9872d9377a17797cfcf6c2fbcee649e3657

### DIFF
--- a/scripts/vulkaninfo_generator.py
+++ b/scripts/vulkaninfo_generator.py
@@ -187,7 +187,6 @@ class VulkanInfoGenerator(OutputGenerator):
 
         self.extFuncs = OrderedDict()
         self.extTypes = OrderedDict()
-        self.extensions = set()
 
     def beginFile(self, genOpts):
         gen.OutputGenerator.beginFile(self, genOpts)
@@ -201,7 +200,6 @@ class VulkanInfoGenerator(OutputGenerator):
 
         for node in root.find('extensions').findall('extension'):
             ext = VulkanExtension(node)
-            self.extensions.add(ext)
             for item in ext.vktypes:
                 self.extTypes[item] = ext
             for item in ext.vkfuncs:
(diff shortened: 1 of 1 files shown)


scripts/vulkaninfo_generator.py
```diff
@@ -200,7 +200,7 @@ def beginFile(self, genOpts):
                     self.constants[item.get('name')] = item.get('value')
 
         for node in root.find('extensions').findall('extension'):
-            ext = VulkanExtension(node)
+            ext = VulkanExtension(node)
```
jeremyk-lunarg: There seems to be some extra whitespace at the end of this line.

shannon-lunarg:
I am still seeing the same warning with a fresh build directory.
```c++
In file included from /home/shannon/vulkan/Vulkan-Tools/vulkaninfo/vulkaninfo.cpp:29:0:
/home/shannon/vulkan/Vulkan-Tools/vulkaninfo/generated/vulkaninfo.hpp: In function ‘const char* VkResultString(VkResult)’:
/home/shannon/vulkan/Vulkan-Tools/vulkaninfo/generated/vulkaninfo.hpp:463:9: warning: case value ‘1000297000’ not in enumerated type ‘VkResult’ [-Wswitch]
         case (1000297000): return "RESULT_EXT_298_RESERVED_VALUE_0_EXT";
         ^~~~
```

shannon-lunarg:
Wait. I forgot I have to generate again. Whoops. Please hold

Issue #134, opened 2019-02-06
rename winrt directory


jjulianoatnv:
The directory `winrt/` contains the Windows Runtime Installer.

The directory name is confusing. `WinRT` is a thing on modern Windows.  It is a modern replacement for what's commonly referred to as `win32`.  The use of the `winrt` directory name creates confusion.  My first assumption was that it is an attempt to provide access to Vulkan from the WinRT API.

Suggest renaming the directory to something like `windows-runtime-installer`.  This name would be accurate and obvious. It's even the same text as the "title" within winrt/README.md.

jeremyk-lunarg:
Fixed by merge of PR #295.

PR #295, opened 2019-11-22
windows: Change winrt name

jeremyk-lunarg:
Changed the name of the `winrt` directory to
`windows-runtime-installer` as suggested in
issue #134.


lenny-lunarg:
I have no problem with this, but definitely get approval from @shannon-lunarg before merging this because this will require changes to the SDK build scripts.

shannon-lunarg:
Once this is merged I have a branch with changes that will work with this name change.

jeremyk-lunarg:
Sounds good.  Merging now.

Issue #285, opened 2019-10-31
vkcube: MacOS vkcube is squished


KarenGhavam-lunarG:
See SDK issue: https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/issues/189

Is there code that can be put into vkcube to avoid this? Is this unique to Mac or happens on other platforms?

mikew-lunarg:
The field of view is being computed from both the window's height and width, so if the window doesn't have a square aspect ratio, the view is squished.  

To preserve an aspect ratio, use either the height or width to compute the FOV, and use that single FOV for both horizontal and vertical.

jeremyk-lunarg:
Fixed by PR #296

PR #296, opened 2019-11-22
vkcube: Fix Cube Squishing on Resize

jeremyk-lunarg:
Modified viewport creation in vkcube to keep the
aspect ratio square and centered.

This change prevents cube from squishing and warping
when the window is resized.

It now shrinks and grows.


jeremyk-lunarg:
This change is passing CI and after testing it on my MacBook it seems to fix issue #285.
Merging.

PR #300, opened 2019-12-06
vulkaninfo: load user32.dll dynamically

charles-lunarg:
Instead of depending on user32.dll, this library is
now loaded dynamically.

This change is needed to redistribute vulkaninfo.exe
as a part of "universal" windows driverm, by making
vulkaninfo pass the 'ApiValidator' test.

See more here https://docs.microsoft.com/en-us/windows-hardware/drivers/develop/validating-universal-drivers

Author: Kamil Szypryt <kamil.szypryt@intel.com>

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -28,6 +28,19 @@
 
 #include "vulkaninfo.hpp"
 
+#ifdef _WIN32
+// Initialize User32 pointers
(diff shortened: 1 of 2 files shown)


charles-lunarg:
@scyganINTEL can you sign the contributor license agreement?

charles-lunarg:
I added a missing `#ifdef _WIN32` that was causing issues for linux as well as formatted them to comply with the clang-format.

Issue #247, opened 2019-08-21
vulkaninfo: Report old drivers being dropped by the loader


lenny-lunarg:
In today's SI meeting, we agreed on a change to loader behavior that will cause the loader to drop very old drivers when newer ones are present. The loader will log a message when ignoring old drivers so that applications can see what is happening. We we need to make sure that vulkaninfo is capable of reporting this warning.

At the very least, vulkaninfo will need to pass on this warning message so that someone who runs it can see the message. It may also be useful provide some sort of hint that this can be fixed by updating drivers. I'll post a link to the loader PR that intrudces this behavior once I open it (which should happen shortly).

charles-lunarg:
Currently vulkaninfo uses debug_report and should return any loader messages that go through that channel. 
We will need to test this to make sure the messages get through but for now no changes need to be made as far as I can tell.

krOoze:
@lenny-lunarg That is actually pretty weird behavior. Is this documented in [oaderAndLayerInterface.md](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md)?

charles-lunarg:
Closing the issue as vulkaninfo uses the debug report callback in vkCreateInstance, which is where this error should be reported.

> @lenny-lunarg That is actually pretty weird behavior. Is this documented in [LoaderAndLayerInterface.md](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md)?

I don't think it is. But the document is long and easy to miss minor details like that.

PR #305, opened 2019-12-11
vulkaninfo: add missing debug report flags

charles-lunarg:
Previously, the debug report extension only reported ERROR and WARNING
messages. This commit add the INFORMATION and DEBUG flags to the
debug report create info struct, to catch more loader messages.

Change-Id: I5332ec6bed5466aaa1263bcee0372c7e22313ead


jeremyk-lunarg:
This changes causes a large amount of output at the beginning of vulkaninfo.
Is there any particular reason this change was suggested?

charles-lunarg:
This was in response to #247 though possibly INFO isn't required due to the volume of output.

jeremyk-lunarg:
It sounds like this message the loader would send would be classified as a warning which vulkaninfo should already report.
Is the loader message of some other classification?

charles-lunarg:
Looking through the loader code, there is only one salient error for the issue using the info tag, but the amount of other spam makes it necessary. 
This might be useful as an option, but that should be its own pr/issue so I'll close this one.

PR #318, opened 2019-12-24
vulkaninfo: exit on invalid gpu selected

charles-lunarg:
Before when using the --json=X flag, if X was out of range it would
default to the first gpu. Now it emits an error and exits.

Changes to be committed:

Change-Id: I0df1100f7ccc6ca37069f7798f3b620efc9a8bee

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -756,7 +756,8 @@ int main(int argc, char **argv) {
     }
 
     if (selected_gpu >= gpus.size()) {
-        selected_gpu = 0;
+        std::cout << "The selected gpu (" << selected_gpu << ") is not in the valid range of 0 to " << gpus.size() - 1 << ".\n";
+        return 0;
     }
 
     std::vector<std::unique_ptr<Printer>> printers;


vulkaninfo/vulkaninfo.cpp
```diff
@@ -756,7 +756,8 @@ int main(int argc, char **argv) {
     }
 
     if (selected_gpu >= gpus.size()) {
-        selected_gpu = 0;
+        std::cout << "The selected gpu (" << selected_gpu << ") is outside of the valid range. Please select a valid gpu.\n";
```
mikew-lunarg: Could the message include the valid bounds?

vulkaninfo/vulkaninfo.cpp
```diff
@@ -756,7 +756,18 @@ int main(int argc, char **argv) {
     }
 
     if (selected_gpu >= gpus.size()) {
-        selected_gpu = 0;
+        std::cout << "The selected gpu (" << selected_gpu << ") is invalid. Please select a valid gpu.\n";
+        std::cout << "Valid gpus are the following: ";
```
mikew-lunarg: It's a range, correct?  No need for a list, how about:
"GPU is not in the valid range of 0 to X"

charles-lunarg: I did that initially. But "valid range of 0 to 0" looks very silly. At first I handled the 1 gpu case seperately, but the text was awkward.

mikew-lunarg: The main point is to give a clue on a multi-gpu system, what the app expects for selection; e.g.: is it numeric or string, an index or some device ID, is it zero-based or one-based, etc.  If they specify "1" on a "range of 0 to 0", the user probably gets the hint.  It's a bit silly to specify an index on a single-gpu system, wouldn't worry about awkwardness of the message

Just give a quick verbal description of the argument check that failed, with enough info that so user can fix.

mikew-lunarg: dont spend too much time on this.  The important thing is to convey info.  dont need to say "please supply a valid gpu", they already know they've given in invalid answer.

charles-lunarg: I see your point. Amended and simplified it.

PR #319, opened 2020-01-06
vulkaninfo: Make presentation surface names a set

charles-lunarg:
Previously, the printing logic stored surface names in a vector.
By switching to a set, this stops surface names from being listed
multiple times.

Files modified:
	vulkaninfo/vulkaninfo.cpp
	vulkaninfo/vulkaninfo.h

Change-Id: I5bf2acfd773c0fef6d17d21d3a489cc30df72e4d

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -1,7 +1,7 @@
 /*
- * Copyright (c) 2015-2019 The Khronos Group Inc.
- * Copyright (c) 2015-2019 Valve Corporation
- * Copyright (c) 2015-2019 LunarG, Inc.
+ * Copyright (c) 2015-2020 The Khronos Group Inc.
+ * Copyright (c) 2015-2020 Valve Corporation
+ * Copyright (c) 2015-2020 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -159,7 +159,7 @@ void DumpSurfaceCapabilities(Printer &p, AppInstance &inst, AppGpu &gpu, AppSurf
     chain_iterator_surface_capabilities2(p, inst, gpu, surface.surface_capabilities2_khr.pNext);
 }
 
-void DumpSurface(Printer &p, AppInstance &inst, AppGpu &gpu, AppSurface &surface, std::vector<std::string> surface_types) {
+void DumpSurface(Printer &p, AppInstance &inst, AppGpu &gpu, AppSurface &surface, std::set<std::string> surface_types) {
     std::string header;
     if (p.Type() == OutputType::text)
         header = std::string("GPU id : ") + std::to_string(gpu.id) + " (" + gpu.props.deviceName + ")";
@@ -189,8 +189,8 @@ void DumpSurface(Printer &p, AppInstance &inst, AppGpu &gpu, AppSurface &surface
 
 struct SurfaceTypeGroup {
     AppSurface *surface;
-    std::vector<std::string> surface_types;
     AppGpu *gpu;
+    std::set<std::string> surface_types;
 };
 
 bool operator==(AppSurface const &a, AppSurface const &b) {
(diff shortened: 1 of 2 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -159,7 +159,7 @@ void DumpSurfaceCapabilities(Printer &p, AppInstance &inst, AppGpu &gpu, AppSurf
     chain_iterator_surface_capabilities2(p, inst, gpu, surface.surface_capabilities2_khr.pNext);
 }
 
-void DumpSurface(Printer &p, AppInstance &inst, AppGpu &gpu, AppSurface &surface, std::vector<std::string> surface_types) {
+void DumpSurface(Printer &p, AppInstance &inst, AppGpu &gpu, AppSurface &surface, std::unordered_set<std::string> surface_types) {
```
lunarpapillo: I'm a little concerned about how the unordered set works with the loop on surface types in line 174.  If the set is truly unordered and no consistent iteration order is guaranteed, this could mean that the output could be different from run to run where there is no actual change.

I'd like to be able to assume that consecutive runs on exactly the same data should be guaranteed to produce exactly the same results.

vulkaninfo/vulkaninfo.cpp
```diff
@@ -210,14 +210,14 @@ void DumpPresentableSurfaces(Printer &p, AppInstance &inst, const std::vector<st
             auto exists = surface_list.end();
             for (auto it = surface_list.begin(); it != surface_list.end(); it++) {
                 // use custom comparator to check if the surface has the same values
-                if (it->gpu == gpu.get() && *it->surface == *surface.get()) {
+                if (it->gpu == gpu.get() && *(it->surface) == *(surface.get())) {
                     exists = it;
```
lunarpapillo: We can `break` after we find the first (and should be the only) match, yes?

charles-lunarg: Yes, that is a good point. It should be the only match, but better to be safe than assume.

vulkaninfo/vulkaninfo.cpp
charles-lunarg: Right, unordered_set is my goto for performance reasons, but here the ordering of set is more important.

vulkaninfo/vulkaninfo.cpp
```diff
@@ -210,14 +210,15 @@ void DumpPresentableSurfaces(Printer &p, AppInstance &inst, const std::vector<st
             auto exists = surface_list.end();
             for (auto it = surface_list.begin(); it != surface_list.end(); it++) {
                 // use custom comparator to check if the surface has the same values
```
mikew-lunarg: nit: I realize this comment wasn't modified by this PR, but is it a TODO, or meant to describe the following line?  Perhaps delete?

charles-lunarg: Its a comment about why its comparing pointer's for the gpu's but comparing values for the surfaces. 
Since its not obvious at all why that is just from looking at the function, I felt it needed a comment.
But re-reading it, the language could be cleared up to not sound like a TODO

vulkaninfo/vulkaninfo.cpp
```diff
@@ -210,14 +210,15 @@ void DumpPresentableSurfaces(Printer &p, AppInstance &inst, const std::vector<st
...
                     exists = it;
+                    break;
                 }
             }
             if (exists != surface_list.end()) {
-                exists->surface_types.push_back(surface.get()->surface_extension.name);
+                exists->surface_types.insert(surface.get()->surface_extension.name);
             } else {
-                surface_list.push_back({surface.get(), {surface.get()->surface_extension.name}, gpu.get()});
+                surface_list.push_back({surface.get(), gpu.get(), {surface.get()->surface_extension.name}});
```
mikew-lunarg: Is this arg reordering a bugfix separate from fixing the name duplication?  If so, prob merits mention in the PR desc.

charles-lunarg: The SurfaceTypeGroup had arguments in "pointer, vector, pointer" order. It seemed more logical to group the pointers together, which necessitates changing the order of the arguments in the `push_back`.

vulkaninfo/vulkaninfo.cpp
mikew-lunarg: ok.  using the term "comparator" lead me to a TODO for an operator==().  Perhaps "comparison"?  definitely a nit  :-)

vulkaninfo/vulkaninfo.cpp
mikew-lunarg: Fair enough.  SurfaceTypeGroup can take either order without modification?

mikew-lunarg:
Needs copyright 2020 bump

vulkaninfo/vulkaninfo.cpp
charles-lunarg: I should of specified that SurfaceTypeGroup is a POD-like struct, and mistakenly said arguments when I meant member variables. The order does matter, so the `surface_list.push_back(..)` needed updating.

mikew-lunarg: my bad, I misread the diff in SurfaceTypeGroup definition.

vulkaninfo/vulkaninfo.cpp
charles-lunarg: Nits are fine by me. I have made it more clear that its a comment, will push with the copyright changes.

PR #322, opened 2020-01-10
vkcube: Fix present semaphore sync

krOoze:
- simplify away QFOT for sanity
- fix present semaphore sync, so we won't signal potentialy still pending semaphore as per https://github.com/KhronosGroup/Vulkan-Docs/issues/1150

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -297,6 +297,7 @@ typedef struct {
     void *uniform_memory_ptr;
     VkFramebuffer framebuffer;
     VkDescriptorSet descriptor_set;
+    VkSemaphore draw_complete_semaphore;
 } SwapchainImageResources;
 
 struct demo {
@@ -357,7 +358,6 @@ struct demo {
     uint32_t graphics_queue_family_index;
     uint32_t present_queue_family_index;
     VkSemaphore image_acquired_semaphores[FRAME_LAG];
-    VkSemaphore draw_complete_semaphores[FRAME_LAG];
     VkPhysicalDeviceProperties gpu_props;
     VkQueueFamilyProperties *queue_props;
     VkPhysicalDeviceMemoryProperties memory_properties;
@@ -1004,7 +1004,7 @@ static void demo_draw(struct demo *demo) {
     submit_info.commandBufferCount = 1;
     submit_info.pCommandBuffers = &demo->swapchain_image_resources[demo->current_buffer].cmd;
     submit_info.signalSemaphoreCount = 1;
-    submit_info.pSignalSemaphores = &demo->draw_complete_semaphores[demo->frame_index];
+    submit_info.pSignalSemaphores = &demo->swapchain_image_resources[demo->current_buffer].draw_complete_semaphore;
     err = vkQueueSubmit(demo->graphics_queue, 1, &submit_info, demo->fences[demo->frame_index]);
     assert(!err);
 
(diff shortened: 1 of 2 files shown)


jeremyk-lunarg:
@krOoze seems clang-format is complaining and Android is failing to build for you in the travis CI.

krOoze:
Should be fine now.

jzulauf-lunarg:
It's not clear that this change reflects the "best practice" for Vulkan. SHARED ownership resources may be restricted by implementations to "least common denominator" layouts (i.e. only those supported on all owners).  While this certainly simplifies cube, historically any changes that reflect the "toy problem" nature of cube have seen pushed back by reviewers that see it as an exemplar for best practices.

If it's a toy problem, the change is fine.  If it's an exemplar, we'd need to show that this is best practice for sharing render targets with present queues.

@mikew-lunarg @oddhack -- since cube.c is explicitly called out in the spec... the cube charter should be clear, no?

krOoze:
I think there is little point in keeping untestable code. Or are we aware of even a single platform that has a separate present qf?

Anyway, the main dish is the semaphore use. I just didn't want to deal with the QFOT adding complexity to the problem.

Which itself is mostly to poke the hornet's nest TBH. Per the best current interpretation of the spec I have been given, this is how they are supposed to be used. Relevant spec issue is there as old as 2016. If you are relying here on some sync guarantee I do not know, I want you to reveal it and ideally fix the spec.

jzulauf-lunarg:
What is the synchronization bug you believe you are fixing.  The only non-explicit "sync guarantee" that we're depending on is that the draw_complete_semaphores has been waited on before the image_acquired_semaphores is signaled.  But that seems implicit in the following

>If semaphore is not VK_NULL_HANDLE, the semaphore must be unsignaled, with no signal or wait operations pending. It will become signaled when the application can use the image.
> 
> Note
> Use of semaphore allows rendering operations to be recorded and submitted before the presentation engine has completed its use of the image.

If the presentation engine has completed with that image, it must have *started* the last present referenced it, and that means the draw_complete has been waited on.

Now I would be far happier if there was an explicit "sync/access scope" defined in the spec :)

krOoze:
You cannot know draw_complete_semaphore is pending or safe for reuse. Only way to know is if the **same** image as used in vkPresent is returned by vkAcquire (and its semaphore is waited on). Or so I have been told in the link in the OP.

Your code only had two semaphores. And their indexes were also cycled in order. So that would violate above interpretation, and the semaphore could still be pending when used in vkSubmit.

jzulauf-lunarg:
I


 t



I see that the guidance has changed though the URL in the OP doesn't make it obvious. 

TL;DR version:  https://github.com/KhronosGroup/Vulkan-Docs/issues/1150#issuecomment-572228230

> The best advice we can offer at the moment is to keep your window system semaphores separate and have a semaphore per WSI image which you use for vkQueuePresentKHR. Once you've gotten that image back from vkAcquireNextImageKHR and you've waited on the fence from vkAcquireNextImageKHR, it should be safe to re-use the semaphore. Prior to that, it's hard to say for sure. Some implementations may be fine and others may have issues.

I think it would still make more sense to have a change that only implements the recommend approach, as oppose to mixing it with the QFO change.

krOoze:
It would make more sense, as well as more work. The xfer semaphores are probably unsafe too...

Maybe next week...

Tony-LunarG:
Closing as no real consensus has been reached and no activity.

PR #325, opened 2020-01-16
vulkaninfo: Support VK_EXT_tooling_info

charles-lunarg:
Added support for the VK_EXT_tooling_info extension.

Files modified:
	modified:   scripts/vulkaninfo_generator.py
	modified:   vulkaninfo/generated/vulkaninfo.hpp
	modified:   vulkaninfo/vulkaninfo.h

Change-Id: I9fc4e239b37fdff9ecdba4bf947a0b5d57d06745

### DIFF
--- a/vulkaninfo/generated/vulkaninfo.hpp
+++ b/vulkaninfo/generated/vulkaninfo.hpp
@@ -773,6 +773,30 @@ void DumpVkDeviceGroupPresentModeFlagBitsKHR(Printer &p, std::string name, VkDev
     DumpVkDeviceGroupPresentModeFlagsKHR(p, name, value, width);
     p.ObjectEnd();
 }
+void DumpVkToolPurposeFlagsEXT(Printer &p, std::string name, VkToolPurposeFlagBitsEXT value, int width = 0) {
+    if (value == 0) p.PrintElement("None");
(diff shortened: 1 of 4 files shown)


vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -773,6 +773,30 @@ void DumpVkDeviceGroupPresentModeFlagBitsKHR(Printer &p, std::string name, VkDev
     DumpVkDeviceGroupPresentModeFlagsKHR(p, name, value, width);
     p.ObjectEnd();
 }
+void DumpVkToolPurposeFlagsEXT(Printer &p, std::string name, VkToolPurposeFlagBitsEXT value, int width = 0) {
+    if (value == 0) p.PrintElement("None");
+    if (1 & value) p.SetAsType().PrintElement("TOOL_PURPOSE_VALIDATION_BIT_EXT");
```
mark-lunarg: Any reason for not using the Vulkan enum bit definitions here (i.e., VK_TOOL_PURPOSE_VALIDATION_BIT_EXT) instead of 1, 2,4,  etc.?

charles-lunarg: This code is autogenerated, so there is little reason to change it to the textual version. If there is any changes to the spec, which alters the value, the autogen script will pick up the change and use the correct values.

PR #328, opened 2020-01-21
vulkaninfo: new vkconfig_outptut backend

charles-lunarg:
This commit adds a new backend, full_json, that prints the same information
as the text & html versions. This output will be used in vkconfig.
    
To enable the output, use the `--full-json` flag from the command line.
    
Many parts of the code needed refactoring to handle the json_full output.
Replacing PrintElement with PrintString, requiring StartObject instead of
StartArray, and removing many branches to make the code simpler to read.
    
json_full does not nest json objects within json arrays, thus keeping every
non-leaf node of a json object named.
    
files modified
     - scripts/vulkaninfo_generator.py
     - vulkaninfo/generated/vulkaninfo.hpp
     - vulkaninfo/outputprinter.h
     - vulkaninfo/vulkaninfo.cpp
     - vulkaninfo/vulkaninfo.h
     - vulkaninfo/vulkaninfo.md

    Change-Id: I33d036a75b65942db1ad62b2a1e0c341a2b5e36c

### DIFF
--- a/scripts/vulkaninfo_generator.py
+++ b/scripts/vulkaninfo_generator.py
@@ -25,6 +25,7 @@ import string
 import xml.etree.ElementTree as etree
 import generator as gen
 import operator
+import json
 from collections import namedtuple
 from collections import OrderedDict
 from generator import *
@@ -57,10 +58,10 @@ license_header = '''
  */
 '''
 
-custom_formaters = '''
-std::ostream &operator<<(std::ostream &o, VkConformanceVersion &c) {
-    return o << std::to_string(c.major) << "." << std::to_string(c.minor) << "." << std::to_string(c.subminor) << "."
-             << std::to_string(c.patch);
+custom_formaters = r'''
+void DumpVkConformanceVersion(Printer &p, std::string name, VkConformanceVersion &c, int width = 0) {
+    p.PrintKeyString("conformanceVersion", std::to_string(c.major)+ "." + std::to_string(c.minor) + "." + std::to_string(c.subminor) + "."
+             + std::to_string(c.patch), width);
 }
 
 template <typename T>
@@ -74,17 +75,20 @@ template <typename T>
 std::string to_hex_str(Printer &p, T i) {
     if (p.Type() == OutputType::json)
         return std::to_string(i);
+    else if (p.Type() == OutputType::vkconfig_output)
+        return std::string("\"") + to_hex_str(i) + std::string("\"");
     else
         return to_hex_str(i);
 }
 '''
+
 
 # used in the .cpp code
 structures_to_gen = ['VkExtent3D', 'VkExtent2D', 'VkPhysicalDeviceLimits', 'VkPhysicalDeviceFeatures', 'VkPhysicalDeviceSparseProperties',
                      'VkSurfaceCapabilitiesKHR', 'VkSurfaceFormatKHR', 'VkLayerProperties', 'VkPhysicalDeviceToolPropertiesEXT']
 enums_to_gen = ['VkResult', 'VkFormat', 'VkPresentModeKHR',
                 'VkPhysicalDeviceType', 'VkImageTiling']
-flags_to_gen = ['VkSurfaceTransformFlagsKHR', 'VkCompositeAlphaFlagsKHR',
+flags_to_gen = ['VkSurfaceTransformFlagsKHR', 'VkCompositeAlphaFlagsKHR', 'VkSurfaceCounterFlagsEXT',
                 'VkDeviceGroupPresentModeFlagsKHR', 'VkFormatFeatureFlags', 'VkMemoryPropertyFlags', 'VkMemoryHeapFlags']
 flags_strings_to_gen = ['VkQueueFlags']
 
@@ -92,10 +96,12 @@ struct_short_versions_to_gen = ['VkExtent3D']
 
 struct_comparisons_to_gen = ['VkSurfaceFormatKHR', 'VkSurfaceFormat2KHR', 'VkSurfaceCapabilitiesKHR',
                              'VkSurfaceCapabilities2KHR', 'VkSurfaceCapabilities2EXT']
+# don't generate these structures
+struct_blacklist = ['VkConformanceVersion']
 
 # iostream or custom outputter handles these types
 predefined_types = ['char', 'VkBool32', 'uint32_t', 'uint8_t', 'int32_t',
-                    'float', 'uint64_t', 'size_t', 'VkDeviceSize', 'VkConformanceVersion']
+                    'float', 'uint64_t', 'size_t', 'VkDeviceSize']
 
 # Types that need pNext Chains built. 'extends' is the xml tag used in the structextends member. 'type' can be device, instance, or both
 EXTENSION_CATEGORIES = {'phys_device_props2': {'extends': 'VkPhysicalDeviceProperties2', 'type': 'both'},
@@ -257,14 +263,13 @@ class VulkanInfoGenerator(OutputGenerator):
         for flag in self.flags:
             if flag.name in types_to_gen:
                 for bitmask in (b for b in self.bitmasks if b.name == flag.enum):
-                    out += PrintFlags(flag, bitmask, self)
                     out += PrintBitMask(bitmask, flag.name, self)
 
             if flag.name in flags_strings_to_gen:
                 for bitmask in (b for b in self.bitmasks if b.name == flag.enum):
                     out += PrintBitMaskToString(bitmask, flag.name, self)
 
-        for s in (x for x in self.all_structures if x.name in types_to_gen):
+        for s in (x for x in self.all_structures if x.name in types_to_gen and x.name not in struct_blacklist):
             out += PrintStructure(s, types_to_gen, names_of_structures_to_gen)
 
         out += "pNextChainInfos get_chain_infos() {\n"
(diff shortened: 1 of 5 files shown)


mikew-lunarg:
suggest calling the flag "--vkconfig-private" (or something similarly cryptic), and avoiding the term "json" altogether, to make extremely obvious what the usecase is, and what is _isn't_

No one outside LunarG has any business knowing about the format.

charles-lunarg:
Should the implementation also remove the obvious "json_full"? That way its not stated as using json (even if the implementation uses all the json based functions)

mikew-lunarg:
In all cases, I suggest the implementation change from "json_full" to "vkconfig_private".  

Thinking further, dont even have a "--vkconfig-private" cmdline switch; it's private, might as well leave it undocumented.

charles-lunarg:
Renamed json_full to vkconfig_output, including the commit message and PR title. Removed the documentation for vkconfig_output as well.

mikew-lunarg:
@charles-lunarg I'm freed up, is it time to review or still in flux?

cdgiessen:
There is still the concern of whether it should print to standard out or a file. Richard requested as a file as if any errors occurs, it is impossible to detect since the desired output and errors all go into the same place and overlap. I believe a separate file is best in this situation with that in mind.

Also, I need to do a once over tomorrow for all output formats, make sure none of the other outputs were altered. But that is more a sanity check than any impedance towards code-review.

charles-lunarg:
There is still the concern of whether it should print to standard out or a file. Richard requested as a file as if any errors occurs, it is impossible to detect since the desired output and errors all go into the same place and overlap. I believe a separate file is best in this situation with that in mind.

Also, I need to do a once over tomorrow for all output formats, make sure none of the other outputs were altered. But that is more a sanity check than any impedance towards code-review.

charles-lunarg:
I have made some minor edits to make the output more uniform and confirmed none of the outputs are totally bugged.
And as per the broader discussion, outputting vkconfig_output to a file is the desired choice.
@mikew-lunarg it's ready to be reviewed

vulkaninfo/vulkaninfo.cpp
```diff
@@ -355,8 +354,8 @@ void GpuDumpProps(Printer &p, AppGpu &gpu) {
     p.AddNewline();
 }
 void GpuDumpQueueProps(Printer &p, std::vector<SurfaceExtension> &surfaces, AppQueueFamilyProperties &queue) {
-    p.SetElementIndex(static_cast<int>(queue.queue_index)).SetSubHeader().ObjectStart("queueProperties");
-    if (p.Type() == OutputType::json) {
+    p.SetSubHeader().SetElementIndex(static_cast<int>(queue.queue_index)).ObjectStart("queueProperties");
+    if (p.Type() == OutputType::json || p.Type() == OutputType::vkconfig_output) {
```
mikew-lunarg: Seems like this if-else tree (and the two below) would be more clearly expressed by switch()es.

(Although so many conditionals based on p.Type() throughout the code really suggest a refactoring into subclasses)

vulkaninfo/vulkaninfo.cpp
```diff
@@ -527,25 +540,28 @@ void GpuDumpFeatures(Printer &p, AppGpu &gpu) {
...
 void GpuDumpFormatProperty(Printer &p, VkFormat fmt, VkFormatProperties prop) {
-    if (p.Type() == OutputType::text) {
-        p.ObjectStart("Properties");
-    } else if (p.Type() == OutputType::html) {
-        p.SetTitleAsType().ObjectStart(VkFormatString(fmt));
-    } else if (p.Type() == OutputType::json) {
-        p.ObjectStart("");
-    }
     if (p.Type() == OutputType::html || p.Type() == OutputType::text) {
+        if (p.Type() == OutputType::html)
```
mikew-lunarg: Another opportunity for separate cases in a switch

vulkaninfo/vulkaninfo.cpp
```diff
@@ -738,7 +757,10 @@ int main(int argc, char **argv) {
 
     // Combinations of output: html only, html AND json, json only, human readable only
     for (int i = 1; i < argc; ++i) {
-        if (strncmp("--json", argv[i], 6) == 0 || strcmp(argv[i], "-j") == 0) {
+        if (strcmp(argv[i], "--vkconfig_output") == 0) {
+            human_readable_output = false;
```
mikew-lunarg: suggest a comment here so it's clear this is intentionally private: e.g.:
"A internal-use-only format for communication with the Vulkan Configurator tool"

vulkaninfo/vulkaninfo.h
```diff
@@ -74,6 +74,7 @@
 bool human_readable_output = true;
 bool html_output = false;
 bool json_output = false;
+bool vkconfig_output = false;
```
mikew-lunarg: Should these bools be private inside main()?  They only seem to have utility between arg parsing and construction of the Printer instance, which all happens in main()

scripts/vulkaninfo_generator.py
```diff
@@ -57,10 +58,10 @@
  */
 '''
 
-custom_formaters = '''
-std::ostream &operator<<(std::ostream &o, VkConformanceVersion &c) {
-    return o << std::to_string(c.major) << "." << std::to_string(c.minor) << "." << std::to_string(c.subminor) << "."
-             << std::to_string(c.patch);
+custom_formaters = r'''
+void DumpVkConformanceVersion(Printer &p, std::string name, VkConformanceVersion &c, int width = 0) {
```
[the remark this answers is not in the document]
charles-lunarg: This is a required change to make vkconfig_output work due to the json nature of the output. Otherwise the conformance version would be an unquoted string and not be valid json.

vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -1186,7 +1325,7 @@ void DumpVkPhysicalDeviceVulkan12Properties(Printer &p, std::string name, VkPhys
     DumpVkDriverId(p, "driverID", obj.driverID, 52);
```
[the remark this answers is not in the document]
charles-lunarg: That is an autogenerated constant for aligning the key and value columns in the output. While the values shouldn't be changing all the time, its not something that needs to be pulled into its own dedicated autogenerated enum

vulkaninfo/outputprinter.h
```diff
@@ -281,21 +295,34 @@ class Printer {
                     out << "[<span class='val'>" << element_index << "</span>]";
                     element_index = -1;
                 }
+                if (count_subobjects >= 0) {
+                    out << ": count = <span class='val'>" << std::to_string(count_subobjects) << "</span>";
+                }
                 out << "</summary>\n";
                 break;
             case (OutputType::json):
+            case (OutputType::vkconfig_output):
```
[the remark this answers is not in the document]
charles-lunarg: This was done to not duplicate the json handling, but there isn't a strong reason to make the switch more complicated with nested-if's. Will revert it in the next revision.

vulkaninfo/generated/vulkaninfo.hpp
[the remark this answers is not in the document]
mikew-lunarg: Suggest putting all auto-generated files in separate commits and clearly identify in the commit description, to help reviews know what can reasonably be skipped.

vulkaninfo/vulkaninfo.h
charles-lunarg: This is because a macro used the `human_readable_output` bool in the header file itself. I could move the json/html/vkconfig bools elsewhere, but that only makes the placement of `human_readable_output` even more confusing than declaring them all together.

vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -596,205 +619,321 @@ std::string VkQueueFlagsString(VkQueueFlags value, int width = 0) {
...
+    std::vector<const char *> strings;
+    if (value == 0) strings.push_back("None");
+    if (1 & value) strings.push_back("SHADER_STAGE_VERTEX_BIT");
+    if (2 & value) strings.push_back("SHADER_STAGE_TESSELLATION_CONTROL_BIT");
+    if (4 & value) strings.push_back("SHADER_STAGE_TESSELLATION_EVALUATION_BIT");
+    if (8 & value) strings.push_back("SHADER_STAGE_GEOMETRY_BIT");
+    if (16 & value) strings.push_back("SHADER_STAGE_FRAGMENT_BIT");
+    if (32 & value) strings.push_back("SHADER_STAGE_COMPUTE_BIT");
+    if (0x0000001F & value) strings.push_back("SHADER_STAGE_ALL_GRAPHICS");
+    if (0x7FFFFFFF & value) strings.push_back("SHADER_STAGE_ALL");
```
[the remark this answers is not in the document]
charles-lunarg: I would rather follow the order that they appear in the spec. That way the order they are listed is consistent and not based off the value, which is less important that the name attached to it.

vulkaninfo/vulkaninfo.md
```diff
@@ -60,6 +60,7 @@ OPTIONS:
                       specifying the gpu-number associated with the gpu of
                       interest. This number can be determined by running
                       vulkaninfo without any options specified.
+--full-json           Produce a json version of all of vulkaninfo to standard output.
```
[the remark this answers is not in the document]
charles-lunarg: This feature is no longer apart of the public api, so I would hesitate adding it to the docs and `--help` text

vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -479,98 +473,127 @@ static const char *VkPresentModeKHRString(VkPresentModeKHR value) {
...
+    if (64 & value) strings.push_back("FORMAT_FEATURE_VERTEX_BUFFER_BIT");
+    if (128 & value) strings.push_back("FORMAT_FEATURE_COLOR_ATTACHMENT_BIT");
+    if (256 & value) strings.push_back("FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT");
+    if (512 & value) strings.push_back("FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT");
+    if (1024 & value) strings.push_back("FORMAT_FEATURE_BLIT_SRC_BIT");
+    if (2048 & value) strings.push_back("FORMAT_FEATURE_BLIT_DST_BIT");
+    if (4096 & value) strings.push_back("FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT");
+    if (16384 & value) strings.push_back("FORMAT_FEATURE_TRANSFER_SRC_BIT");
+    if (32768 & value) strings.push_back("FORMAT_FEATURE_TRANSFER_DST_BIT");
+    if (131072 & value) strings.push_back("FORMAT_FEATURE_MIDPOINT_CHROMA_SAMPLES_BIT");
```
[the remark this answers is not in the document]
charles-lunarg: Added in the next revision

vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -596,205 +619,321 @@ std::string VkQueueFlagsString(VkQueueFlags value, int width = 0) {
...
 void DumpVkResolveModeFlags(Printer &p, std::string name, VkResolveModeFlags value, int width = 0) {
     if (p.Type() == OutputType::json) { p.PrintKeyValue(name, value); return; }
-    p.ObjectStart(name);
-    DumpVkResolveModeFlags(p, name, static_cast<VkResolveModeFlagBits>(value), width);
-    p.ObjectEnd();
+    auto strings = VkResolveModeFlagBitsGetStrings(static_cast<VkResolveModeFlagBits>(value));
+    if (static_cast<VkResolveModeFlagBits>(value) == 0) {
+        p.ArrayStart(name, 0);
+        p.SetAsType().PrintString("None");
+        p.ArrayEnd();
```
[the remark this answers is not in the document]
charles-lunarg: Added and applied liberally.

vulkaninfo/vulkaninfo.cpp
charles-lunarg: Most of these if-else nests were just from json schema handling. I moved all the json dumping logic into their own functions and make the entire codebase much simpler to read

vulkaninfo/vulkaninfo.cpp
charles-lunarg: Added

vulkaninfo/vulkaninfo.cpp
```diff
@@ -807,17 +845,24 @@ int main(int argc, char **argv) {
     if (json_output) {
         printers.push_back(std::unique_ptr<Printer>(new Printer(OutputType::json, out, selected_gpu, instance.vk_version)));
     }
+    if (vkconfig_output) {
+        vkconfig_out = std::ofstream("vulkaninfo.json");
```
mikew-lunarg: nit: Does the consumer library of this file require the ".json" extension?  If not, consider a non-descriptive "vulkaninfo.vkconfig" filename to dissuade any reverse engineers .

What is the lifetime of the file?  Who is responsible for removing the file?

charles-lunarg: vkconfig will be responsible for the file. Right now it prints to the current directory, but a "output_path" command line argument should be added to facilitate putting `vulkaninfo.json` in a temp directory. This should be its own PR, as this one is too big already.

mikew-lunarg: Be nice not to call it .json; perhaps .bin or .vkconfig, again to emphasize its specific purpose.

the undocumented --vkconfig_output could take an argument for the path.

charles-lunarg: I will leave it as .json for the time being to stay consistent with via.
And adding a path parameter after --vkconfig_output is not hard so I will add it to this commit so there won't be multiple versions of vulkaninfo with different behavior floating out there in the wild.

PR #329, opened 2020-01-22
vulkaninfo: use a helper for the two-call idiom

Ralith:
Drastically reduces repetition and ensures consistent error handling.


jeremyk-lunarg:
Looks like AppVeyor had some issues and your changes need to be reformatted to make travis-ci happy.

I would suggest running clang-format over your code and push up the modifications it makes.

PR #334, opened 2020-01-24
vulkaninfo: use gpu version instead of instance

charles-lunarg:
vulkaninfo would previously use the instance version for determining
what capabilities to display. This leads to issues when running a 1.2
loader on a 1.1 or 1.0 driver.

Changes to be committed:
	modified:   vulkaninfo/vulkaninfo.h

Change-Id: I7f12c10d48f99d429025337b00c677936a040d13

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -349,7 +349,7 @@ void GpuDumpProps(Printer &p, AppGpu &gpu) {
     if (p.Type() != OutputType::json) {
         if (gpu.inst.CheckExtensionEnabled(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME)) {
             void *place = gpu.props2.pNext;
-            chain_iterator_phys_device_props2(p, gpu.inst, gpu, place, gpu.inst.vk_version);
+            chain_iterator_phys_device_props2(p, gpu.inst, gpu, place, gpu.api_version);
         }
     }
     p.AddNewline();
@@ -521,7 +521,7 @@ void GpuDumpFeatures(Printer &p, AppGpu &gpu) {
     if (p.Type() != OutputType::json) {
         if (gpu.inst.CheckExtensionEnabled(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME)) {
             void *place = gpu.features2.pNext;
-            chain_iterator_phys_device_features2(p, gpu, place, gpu.inst.vk_version);
+            chain_iterator_phys_device_features2(p, gpu, place, gpu.api_version);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -349,7 +349,7 @@ void GpuDumpProps(Printer &p, AppGpu &gpu) {
     if (p.Type() != OutputType::json) {
         if (gpu.inst.CheckExtensionEnabled(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME)) {
             void *place = gpu.props2.pNext;
-            chain_iterator_phys_device_props2(p, gpu.inst, gpu, place, gpu.inst.vk_version);
+            chain_iterator_phys_device_props2(p, gpu.inst, gpu, place, gpu.api_version);
```
lunarpapillo: Shouldn't this be the smaller of the loader/API version and the GPU/driver version?

Using just the API version will do bad things if you have an early loader and a more modern driver.  (Maybe this isn't an expected use case?)

mikew-lunarg:
Agreed on supporting all three.

charles-lunarg:
Regarding specific situations
* loader is the same version as the driver
Already handled

* loader is later than the driver
The current PR fixes this issue and shouldn't cause any issues. 

* loader is earlier than the driver
Would be handled by using the loader version if its less than the driver version.

Proposed mechanism to handle newer drivers with older loaders
```cpp
uint32_t version_major = VK_VERSION_MAJOR(props.apiVersion);
uint32_t version_minor = VK_VERSION_MINOR(props.apiVersion);
uint32_t version_patch = VK_VERSION_PATCH(props.apiVersion);

api_version = {inst.vk_version.major < version_major ? inst.vk_version.major : version_major,
              inst.vk_version.minor < version_minor ? inst.vk_version.minor : version_minor,
              inst.vk_version.patch < version_patch ? inst.vk_version.patch : version_patch};
```

lunarpapillo:
> ```c++
> api_version = {inst.vk_version.major < version_major ? inst.vk_version.major : version_major,
>               inst.vk_version.minor < version_minor ? inst.vk_version.minor : version_minor,
>               inst.vk_version.patch < version_patch ? inst.vk_version.patch : version_patch};
> ```

This comparison won't work in all cases.  Consider the results of the above algorithm with a pair of input versions **1.13.172** and **3.2.393**; the above algorithm calculates the minimum to be 1.2.172, which is not correct (the correct answer is clearly 1.13.172).

I recommend just doing a straight comparison/minimum of the two aggregate version numbers.  The way the aggregate version number is created:
```c++
#define VK_MAKE_VERSION(major, minor, patch) \
    (((major) << 22) | ((minor) << 12) | (patch))
```
ensures that the two will be correctly comparable, unless `patch` overflows 12 bits,  `minor` overflows 10 bits, or `major` exceeds 9 bits (which could make it negative if the version is a 32-bit signed integer).

charles-lunarg:
I see the issue. The entire value needs to be compared, not individual parts. The amended version I just pushed does that so creating a new minimum of the components of major.minor.patch is no longer possible.

PR #339, opened 2020-01-31
build: Update known-good for 1.2.132 header

shannon-lunarg:


shannon-lunarg:
@jeremyk-lunarg Can you verify this works on macOS?

Issue #342, opened 2020-02-10
Vulkaninfo dereferences uninitialized pointer if extension is not enabled


Tony-LunarG:
If a driver doesn't support VK_KHR_get_surface_capabilities2, initialization of surface_capabilities2_khr and surface_capabilities2_ext is correctly skipped, but then they are used in DumpSurfaceCapabilities, leading to dereferencing an uninitialized pointer.

PR #343, opened 2020-02-10
vulkaninfo: Check for surface extensions

charles-lunarg:
Previously, vulkaninfo assumed the surface counter and surface capabilities 2 extensions were
supported. Now it correctly checks for them.

Files modified:
	vulkaninfo/vulkaninfo.cpp

Change-Id: Ia5a8e22fd062f555d60ca0ebf27e06afbb1bb8cc


charles-lunarg:
fixes #342

PR #345, opened 2020-02-18
Vulkaninfo: dynamically load vulkan-1.dll

charles-lunarg:
This changes vulkaninfo to not statically depend on vulkan-1.dll
to satisfy Microsoft ApiValidator requirements.

volk (https://github.com/zeux/volk) is imported and used in this
commit to facilitate dynamic loading of entrypoints

Author: Kamil Szypryt <Kamil.Szypryt@intel.com>
Change-Id: I74e84bef5642071e855e0f1f9d80e707d46d91df

### DIFF
--- a/vulkaninfo/CMakeLists.txt
+++ b/vulkaninfo/CMakeLists.txt
@@ -34,18 +34,24 @@ if(PYTHONINTERP_FOUND)
 endif()
 
 if(WIN32)
-    add_executable(vulkaninfo vulkaninfo.cpp vulkaninfo.rc)
+    add_executable(vulkaninfo vulkaninfo.cpp vulkaninfo.rc volk/volk.c)
 elseif(APPLE)
     add_executable(vulkaninfo
                    vulkaninfo.cpp
                    ${CMAKE_CURRENT_SOURCE_DIR}/macOS/vulkaninfo/metal_view.mm
-                   ${CMAKE_CURRENT_SOURCE_DIR}/macOS/vulkaninfo/metal_view.h)
+                   ${CMAKE_CURRENT_SOURCE_DIR}/macOS/vulkaninfo/metal_view.h
+                   volk/volk.c)
 else()
-    add_executable(vulkaninfo vulkaninfo.cpp)
+    add_executable(vulkaninfo vulkaninfo.cpp volk/volk.c)
 endif()
 
 target_include_directories(vulkaninfo PRIVATE ${CMAKE_SOURCE_DIR}/vulkaninfo)
 target_include_directories(vulkaninfo PRIVATE ${CMAKE_SOURCE_DIR}/vulkaninfo/generated)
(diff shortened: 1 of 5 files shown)


jeremyk-lunarg:
This PR seems to include a third-party dependency.  Are we sure we want to add it?

charles-lunarg:
The first version of this commit manually loaded the .dll and then used vulkan.hpp's dispatch table mechanism to get the function pointers.
Volk does both and requires minimal code changes to vulkaninfo.h. 
Updaing Volk would only be required when supporting new functions the headers add which are not autogenerated. Currently, the majority of extensions only need vulkaninfo-autogeneration and does not require Volk changes.

vulkaninfo/CMakeLists.txt
```diff
@@ -57,21 +63,21 @@ if(UNIX AND NOT APPLE) # i.e. Linux
         find_package(XCB REQUIRED)
         target_include_directories(vulkaninfo PRIVATE ${XCB_INCLUDE_DIRS})
         target_link_libraries(vulkaninfo ${XCB_LIBRARIES})
-        target_compile_definitions(vulkaninfo PRIVATE -DVK_USE_PLATFORM_XCB_KHR)
+        target_compile_definitions(vulkaninfo PRIVATE -DVK_USE_PLATFORM_XCB_KHR -DVK_NO_PROTOTYPES)
```
mikew-lunarg: Is this (and below) part of using volk, or some other maintenance?

vulkaninfo/volk/LICENSE.md
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2018-2019 Arseny Kapoulkine
```
mikew-lunarg: The volk-adding commit should also include details on exactly which URL and version/tag/commit was brought over from volk's repo

vulkaninfo/CMakeLists.txt
charles-lunarg: `DVK_NO_PROTOTYPES` is whats needed for stopping the headers from adding the function prototypes. This is required so vulkaninfo uses the function pointers provided by volk.

mikew-lunarg: ok

charles-lunarg:
It was not known at the time of this PR that Volk was being added to Vulkan-Tools in general. 
If that is the plan, then this commit needs to be split up and Volk added seperately.

mikew-lunarg:
I believe there was survey feedback suggesting the addition of volk, ask @KarenGhavam-lunarG 

Regardless, volk is a general purpose library, and should not have an instance private to vulkaninfo.  And proper version/tag tracking is needed.

mikew-lunarg:
Per offline discussion, we need to officially decide how to integrate volk into SDK tools.

jeremyk-lunarg:
Discussed this in a meeting.
Decided on a solution that does not use additional external dependencies.
Closing.

PR #350, opened 2020-02-27
vulkaninfo: Refactor error handling

charles-lunarg:
Previously, if vulkaninfo encountered any errors, it shut down then
and there. This is bad for json output where you are left with an
invalid file. By using exceptions, vulkaninfo will catch errors,
print it to cerr, then finish off the file so its left in a valid
state.

Change-Id: I2c0116b435713df6135b9c606fa95cfadbfdf8d8

### DIFF
--- a/vulkaninfo/outputprinter.h
+++ b/vulkaninfo/outputprinter.h
@@ -181,12 +181,14 @@ class Printer {
                 out << "\t}";
                 indents++;
                 is_first_item.push(false);
+                is_array.push(false);
                 break;
             case (OutputType::vkconfig_output):
                 out << "{\n";
                 out << "\t\"Vulkan Instance Version\": \"" << VkVersionString(vulkan_version) << "\"";
                 indents++;
                 is_first_item.push(false);
+                is_array.push(false);
                 break;
             default:
                 break;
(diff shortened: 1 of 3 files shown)


mikew-lunarg:
Couple thoughts from _very_ quick skim

1. Seems like should be several commits, for easier review?  The changes seem to split along semantic groups.
1. What is the driving requirement?  Is there actually a requirement to deliver parsable output on an error exit?  Are errors that common?  Is parsable incomplete data risking being misinterpreted as a valid result?  If the app returns an error code, a downstream consumer can still deliver a message, and perhaps display the output as unstructured text for diagnosis.

charles-lunarg:
1. Yes, there are several groups of fixes, namely the IndentWrapper and EXIT_ERR replacement with THROW_ERR. The issue was that after writing the THROW_ERR to use an exception, the Indentation wasn't being respected (as the exception bypassed any `p.Indent()` calls). The IndentWrapper then is needed to fix the output and keep everything working as intended.

2. The need is to allow VkConfig to display incomplete Json files while still being able to report the issue to the user, instead of only being able to display the error. HTML had no issue since a malformed html document will still be drawable, to a certain extent. Json requires a valid document, so this commit makes it so that errors should be gracefully handled.

charles-lunarg:
I agree, the IndentWrapper is backwards from intuition. Added a bit of documentation, as well as documented many of the other additions made in the PR.

PR #363, opened 2020-04-02
cube: Pick non-SRGB surface formats first

Joshua-Ashton:
Pick some common non-SRGB formats first. Some drivers such as RADV expose SRGB formats first which causes incorrect rendering otherwise.


jeremyk-lunarg:
@Joshua-Ashton, looks like there are some problems with formatting.
Please run clang-format on your changes and update the commit.

Joshua-Ashton:
I forced pushed here and it seems to have passed the tests.

PR #364, opened 2020-04-07
vulkaninfo: fix app bundle not finding loader

charles-lunarg:
The change of vulkaninfo to dlopen the loader broke the app bundle.
This commit fixes it by manually setting up DYLD_LIBRARY_PATH before
starting vulkaninfo, ensuring that it can be found.


jeremyk-lunarg:
Tried building and running this on my MacBook.
Fails with the following error:
```
LSOpenURLsWithRole() failed with error -10810 for the file [USER DIRECTORY]/Repos/Vulkan-Tools/build/install/vulkaninfo/vulkaninfo.app.
```

Looking in the bundle, it seems CMake isn't including `vulkaninfo_run.command` even in the `build/install` directory.

charles-lunarg:
Ack, that should of been copied into the build directory.
Fixed & force-pushed.

PR #365, opened 2020-04-07
Make Mock ICD physical device static

mark-lunarg:
PDev created at EnumPhysDevices time, but destroyed at DestroyInstance-time. This crashes when mixing Create/DestroyInstance calls as EnumPhysDevices is not required to be called.

### DIFF
--- a/icd/generated/mock_icd.cpp
+++ b/icd/generated/mock_icd.cpp
@@ -31,7 +31,7 @@ using std::unordered_map;
 // Map device memory handle to any mapped allocations that we'll need to free on unmap
 static unordered_map<VkDeviceMemory, std::vector<void*>> mapped_memory_map;
 
-static VkPhysicalDevice physical_device = nullptr;
+static VkPhysicalDevice physical_device = (VkPhysicalDevice)CreateDispObjHandle();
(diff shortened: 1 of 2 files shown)


icd/generated/mock_icd.cpp
```diff
@@ -31,7 +31,7 @@ using std::unordered_map;
 // Map device memory handle to any mapped allocations that we'll need to free on unmap
 static unordered_map<VkDeviceMemory, std::vector<void*>> mapped_memory_map;
 
-static VkPhysicalDevice physical_device = nullptr;
+static VkPhysicalDevice physical_device = (VkPhysicalDevice)CreateDispObjHandle();
```
krOoze, line 34: Hm, normally physical devices of distinct instances should have disctinct handle (=pointer) value. Maybe this is asking for more problems?

Why not just `if` the destruction in `vkDestroyInstance`?

mark-lunarg: The behavior is unchanged from the original design, a single PDev, which seems fine for how the mock ICD is used, though it could cause trouble in a pathological case.  I don't understand "'if' the destruction" -- the PDev should not be destroyed as long as there other instances exist, which'd mean tracking the instances, but maybe I misunderstood?

krOoze: Ah, right, it is `static` anyway.

PR #368, opened 2020-04-08
icd: Fix vkGetSwapchainImages behavior

krOoze:
- add missing lock
- return correct values via return and count
- bump to two images
- make it return the same image handles for the same swapchain


mark-lunarg:
@krOoze, had to revert the "Fix vkGetSwapchainImages behavior commit" along with a commit in the companion PR -- these caused the "VkLayerTest.SwapchainAcquireTooManyImages2KHR" test in VVL to fail which blocked CI.  If you can take a look, and figure out what's up we can get them pushed back ASAP.

krOoze:
Hm, also passes for me.

krOoze:
Ah, they fail when both together.

krOoze:
@mark-lunarg This one should be OK. Problem with #369 is that it is returning the same image without it being given back by vkQueuePresent with combination of layers accounting of acquire count being an  individual image state. Quick fix should be to revert this one to one image.

krOoze:
@mark-lunarg Fix in #372

PR #371, opened 2020-04-08
mockicd:Track memory allocation size

tobine:
Save the size of memory allocations. If a vkMapMemory is called on a
known allocation with VK_WHOLE_SIZE, then map based on the actual size
of the allocation.


tobine:
Clean on farm. This should improve some testing where memory is mapped as mapping of whole allocation is now based on actual size of the initially allocated allocation instead of hard-coded.

PR #372, opened 2020-04-10
icd: Fix vkGetSwapchainImages behavior reup

krOoze:
Reup of #368 

- add missing lock
- return correct values via return and count
- <s>bump to two images</s>
- make it return the same image handles for the same swapchain


mark-lunarg:
Thanks Petr!  This works fine with Master.

PR #374, opened 2020-04-13
build: Remove glslang dependancy from known-good

jeremyk-lunarg:
Removed glslang dependancy from known-good.json


jeremyk-lunarg:
This is passing all CI.
If no one objects by 4:00pm today, I'll merge it.

PR #375, opened 2020-04-14
icd: Make VkPhysicalDevice unique

krOoze:
It helps tools that imbue the `VkPhysicalDevice` with other state via a `std::unordered_map`, such as the devsim layer.

### DIFF
--- a/icd/generated/mock_icd.cpp
+++ b/icd/generated/mock_icd.cpp
@@ -22,6 +22,7 @@
 #include "mock_icd.h"
 #include <stdlib.h>
 #include <algorithm>
+#include <array>
 #include <vector>
 #include "vk_typemap_helper.h"
 namespace vkmock {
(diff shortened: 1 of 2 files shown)


icd/generated/mock_icd.cpp
```diff
@@ -193,20 +198,28 @@ static VKAPI_ATTR void VKAPI_CALL DestroyInstance(
     const VkAllocationCallbacks*                pAllocator)
 {
 
-    DestroyDispObjHandle((void*)instance);
+    if (instance) {
+        for (const auto physical_device : physical_device_map.at(instance))
+            DestroyDispObjHandle((void*)physical_device);
+        DestroyDispObjHandle((void*)instance);
```
mikes-lunarg, line 204: Do you want to erase the `physical_device_map` entry once you are done with it?

krOoze: dammit

PR #376, opened 2020-04-16
Give a reasonable image memory size. Every texel size is 4 bytes for now.

locke-lunarg:


icd/generated/mock_icd.cpp
```diff
@@ -849,6 +856,39 @@ static VKAPI_ATTR VkResult VKAPI_CALL CreateImage(
...
+        case VK_FORMAT_G12X4_B12X4R12X4_2PLANE_420_UNORM_3PACK16:
+        case VK_FORMAT_G12X4_B12X4R12X4_2PLANE_422_UNORM_3PACK16:
+        case VK_FORMAT_G16_B16R16_2PLANE_420_UNORM:
+        case VK_FORMAT_G16_B16R16_2PLANE_422_UNORM:
+            planeCount = 2;
+            break;
+        default:
+            break;
+    }
+    image_memory_size_map[device][*pImage] *= planeCount;
```
tobine, line 891: Note that you could put this inline for planeCount 2 or 3 above and avoid this extra mul in the common case of planeCount == 1. Saving individual instructions isn't a concern for mock ICD so I'm fine with it this way.

icd/generated/mock_icd.cpp
```diff
@@ -41,6 +41,7 @@ static unordered_map<VkDeviceMemory, VkDeviceSize> allocated_memory_size_map;
 
 static unordered_map<VkDevice, unordered_map<uint32_t, unordered_map<uint32_t, VkQueue>>> queue_map;
 static unordered_map<VkDevice, unordered_map<VkBuffer, VkBufferCreateInfo>> buffer_map;
+static unordered_map<VkDevice, unordered_map<VkImage, VkDeviceSize>> image_memory_size_map;
```
tobine, line 44: When the device is destroyed we should remove any corresponding entry from this map

locke-lunarg: Fixed

icd/generated/mock_icd.cpp
locke-lunarg: Fixed

PR #377, opened 2020-04-17
icd: Set texel size is 32

locke-lunarg:
4 bytes is  too small for many format. It causes fail in some situation.
32 bytes is the biggest size in the whole format, so it should be better.

### DIFF
--- a/icd/generated/mock_icd.cpp
+++ b/icd/generated/mock_icd.cpp
@@ -858,9 +858,9 @@ static VKAPI_ATTR VkResult VKAPI_CALL CreateImage(
 {
     unique_lock_t lock(global_lock);
     *pImage = (VkImage)global_unique_handle++;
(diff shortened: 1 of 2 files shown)


icd/generated/mock_icd.cpp
```diff
@@ -858,9 +858,9 @@ static VKAPI_ATTR VkResult VKAPI_CALL CreateImage(
 {
     unique_lock_t lock(global_lock);
     *pImage = (VkImage)global_unique_handle++;
-    // TODO: A texel size is 4 bytes for now whatever the format is. It could be changed to more accurate size if need be.
+    // TODO: A texel size is 32 bytes. That is the biggest size in the whole of formats.  It could be changed to more accurate size if need be.
```
tobine, line 861: Suggest re-wording 2nd sentence. "This accounts for the largest possible pixel size of any format."

PR #384, opened 2020-04-26
vulkaninfo: use generate_source.py for autogen

### LINKED ISSUE #356: VulkanInfo generated files should be incorporated into `generate_source.py`
These should work like the other generated files and be checked into the repo.

charles-lunarg:
Previously, to update the autogen for vulkaninfo required running a
separate cmake target. This commit puts it all into the same target
for ease of maintenance.

Fixes #356 

Change-Id: I98e35b01ee164e9917564f6b603e4a78c6138041

### DIFF
--- a/scripts/generate_source.py
+++ b/scripts/generate_source.py
@@ -39,52 +39,72 @@ def main(argv):
     group.add_argument('-v', '--verify', action='store_true', help='verify repo files match generator output')
     args = parser.parse_args(argv)
 
-    gen_cmds = [[common_codegen.repo_relative('scripts/kvt_genvk.py'),
-                 '-registry', os.path.abspath(os.path.join(args.registry,  'vk.xml')),
-                 '-quiet',
-                 filename] for filename in ['vk_typemap_helper.h',
+    # output paths and the list of files in the path
+    files_to_gen = {str(os.path.join('icd','generated')) : ['vk_typemap_helper.h',
                                             'mock_icd.h',
-                                            'mock_icd.cpp']]
+                                            'mock_icd.cpp'],
+                    str(os.path.join('vulkaninfo','generated')): ['vulkaninfo.hpp']}
 
-    repo_dir = common_codegen.repo_relative('icd/generated')
+    #base directory for the source repository
+    repo_dir = common_codegen.repo_relative('')
 
-    # get directory where generators will run
+    # get directory where generators will run if needed
     if args.verify or args.incremental:
         # generate in temp directory so we can compare or copy later
         temp_obj = tempfile.TemporaryDirectory(prefix='VulkanLoader_generated_source_')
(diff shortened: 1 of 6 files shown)


scripts/generate_source.py
```diff
@@ -39,39 +39,56 @@ def main(argv):
...
+                return 1
+        
     # optional post-generation steps
     if args.verify:
         # compare contents of temp dir and repo
-        temp_files = set(os.listdir(temp_dir))
-        repo_files = set(os.listdir(repo_dir))
+        temp_files = set()
+        for path in files_to_gen.keys():
+            temp_files.union(set(os.listdir(os.path.join(temp_dir, path))))
```
mikes-lunarg, line 86: The result of this union operation are not used, temp_files and repo_files remain empty and the compare step ends up being a no-op

I switched it to from `union` to `update` and --verify still fails because the stored path is missing the intermediate directory (`"icd/generated"` and `"vulkaninfo/generated"`)

charles-lunarg: You're completely right. Seems I saw the 'success' and didn't think twice about how nothing was actually being compared. 
I'll fix that and amend the commit.

charles-lunarg: It should work now as intended, can you confirm?

mikes-lunarg:
CI looks mad, for windows I think you will have to rebase against master and updated the .gitattributes file for the vulkaninfo generated dir because of the line-endings change we absorbed with the 140 header

charles-lunarg:
Ah, thats what I need to do to fix it. 
But because of the failing CI, I found 2 places that used `/` path separators, so now windows should have the right paths regardless.

mikes-lunarg:
I'm pretty sure the issue with AppVeyor is line-endings. First it does a clone with master and gets a CRLF version of vulkaninfo.hpp. Then it switches to your branch which has no changes to vulkaninfo.hpp so the CRLF line endings persist, despite the .gitattributes specifying LF.

3 Options:
- Create a separate PR for the .gitattributes, merge it first, then rebase this change
- Introduce a change to vulkaninfo.hpp in your branch
- Just merge it as-is, it _should_ be fixed once it hits master

mikew-lunarg:
I'd definitely suggest a separate PR for .gitattributes.  Git gets very confused when the .gitattributes file comes and goes, so it's wise to also rebase any outstanding branches on top of the .gitattributes addition, so it stays in existence.

Ideally, .gitattributes is added as the very first commit in a repo.

charles-lunarg:
I added a space to the header in the autogen. That seems to have done it, thanks! 

The change to the git attributes is to add the missing details for the auto-generated vulkaninfo file, so it should only affect that file. Should that necessitate a new commit? Since I rebuilt the autogen file, this commit should also have the needed changes.

PR #390, opened 2020-04-28
vulkaninfo: Meet extension version dependencies

krOoze:
Fixes error when VK_KHR_surface_protected_capabilities
is enabled with Vulkan 1.0

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -622,7 +622,7 @@ struct AppInstance {
                                                              DbgCallback};
 
         const VkApplicationInfo app_info = {
-            VK_STRUCTURE_TYPE_APPLICATION_INFO, nullptr, app_short_name, 1, nullptr, 0, VK_API_VERSION_1_0};
+            VK_STRUCTURE_TYPE_APPLICATION_INFO, nullptr, app_short_name, 1, nullptr, 0, instance_version};
 
         AppCompileInstanceExtensionsToEnable();
 


krOoze:
@charles-lunarg It will not be added, because driver that does not support the version must not report the extension.

charles-lunarg:
Ah I had it backwards then. I had made a similar fix the other day in a unpublished branch were I used the solution above, because at the time I was hesitant to change the apiVersion in VkAplicationInfo. However thinking it through, using the version queried is a better solution overall, as you very rightly point out that on a 1.0 platform, that extension should not be reported.

krOoze:
Right. It should be mostly safe to use the queried version. Only things I can think of

- From time to time, there might be contra-dependencies. So far only `VK_AMD_negative_viewport` I think.
- Newer version disables implicit layers that do not support it. No harm done I think; they are either drivers, in which case they should be always updated, or they should not change the info anyway.
- Malicious driver could report different info for different versions. More of a theoretical concern. And no harm done; we do not know anyway for which version the user wants to know the info.

charles-lunarg:
Definitely worthwhile to think about the consequences, but none of those seem likely or needing to be handled. Negative viewport was obsoleted by the VK_KHR_maintenance1, disabling implicit layers shouldn't impact vulkaninfo due to it not enabling any layers in the first place (or rather, no behavior it reports should depend on implicit layers), and worrying about malicious drivers feels like the ultimate form of defensive coding. 

Also, one of the vulkaininfo alterations made for the 1.2 release was for the autogen to support reporting the correct struct names depending on the version or available extensions. That should make sure we report the 'most' correct info possible.

krOoze:
> obsoleted by the VK_KHR_maintenance1

Technically, "deprecated" which means the app must not try to enable it. Should be straightforward to handle, just needs to be kept in mind if Khronos decides to do more of those in future.

krOoze:
NVM, I see you already only enabling Swapchain and Device Group extensions on `vkCreateDevice`, so that is non-issue.

charles-lunarg:
Now just waiting for CI to complete...

PR #393, opened 2020-05-07
build: Update known-good files for 1.2.140 header

mikes-lunarg:
Changes:
- Modify cubepp to use initializer lists for some calls because the
  compiler could not deduce the correct overload from the int+pointer
  array specification
- Propagate genvk.py changes from VVL
- Modify .gitattributes so generated files are checked out with LF line-endings
- Updated known-good files
- Generated new source files


mikes-lunarg:
Forgot the android known-good file, another update incoming

PR #408, opened 2020-06-02
vulkaninfo: fix indent width for present support

charles-lunarg:
Change-Id: I6fafdfac5d508cb02026c4cf1d1dbf7adf866bd4

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -377,7 +377,7 @@ void GpuDumpQueueProps(Printer &p, std::vector<SurfaceExtension> &surfaces, AppQ
     p.PrintKeyValue("timestampValidBits", queue.props.timestampValidBits, 27);
 
     if (queue.is_present_platform_agnostic) {
-        p.PrintKeyString("present support", queue.platforms_support_present ? "true" : "false");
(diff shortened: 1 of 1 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -377,7 +377,7 @@ void GpuDumpQueueProps(Printer &p, std::vector<SurfaceExtension> &surfaces, AppQ
     p.PrintKeyValue("timestampValidBits", queue.props.timestampValidBits, 27);
 
     if (queue.is_present_platform_agnostic) {
-        p.PrintKeyString("present support", queue.platforms_support_present ? "true" : "false");
+        p.PrintKeyString("present support", queue.platforms_support_present ? "true" : "false", 27);
```
jeremyk-lunarg, line 380: I'm seeing the number 27 here and a couple other places without a lot of context.

What is this number for, and would it be cleaner to use a variable for it if it is used in multiple places?

charles-lunarg: Its the indent width to make sure all the ` = value` are lined up vertically. In hindsight, it definitely looks very 'magical' so I should make another commit that uses a named variable in place of the random constant. But that's for a latter commit, not this one.

PR #411, opened 2020-06-04
icd:Correctly Support Vulkan 1.1

tobine:
Mock ICD supports Vulkan 1.1 but was only returning 1.0.
Fixing this.

### DIFF
--- a/scripts/mock_icd_generator.py
+++ b/scripts/mock_icd_generator.py
@@ -496,6 +496,10 @@ CUSTOM_C_INTERCEPTS = {
 'vkEnumerateInstanceLayerProperties': '''
     return VK_SUCCESS;
 ''',
+'vkEnumerateInstanceVersion': '''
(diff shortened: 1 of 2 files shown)


scripts/mock_icd_generator.py
```diff
@@ -765,7 +769,7 @@
 ''',
 'vkGetPhysicalDeviceProperties': '''
     // TODO: Just hard-coding some values for now
-    pProperties->apiVersion = VK_API_VERSION_1_0;
+    pProperties->apiVersion = VK_API_VERSION_1_1;
```
mikew-lunarg, line 772: Might be more robust to use a single local constant for the version, rather than 3 references to the VK constant.

tobine: Thanks, Mike. I'll land this and submit a follow-on to converge all the version refs.

PR #414, opened 2020-06-19
scripts: Use named params for generator options

shannon-lunarg:
These changes are in preparation for the upcoming header update.

When calling in the vulkan header python generator the options have
changed in 1.2.142 which causes the options passed by the tools to be
off by one. This ends up setting the emitExtensions to the value of the
sortProcedure and eventually causes the script to fail.

This CL updates the tools to used named parameters for the generator so
that the addition of the genpath does not throw off the parameter
positioning.

Change-Id: I580d5f07ec3ff9e523dbf738561c2e53cfd42796

### DIFF
--- a/scripts/mock_icd_generator.py
+++ b/scripts/mock_icd_generator.py
@@ -1057,9 +1057,19 @@ class MockICDGeneratorOptions(GeneratorOptions):
                  alignFuncParam = 0,
                  expandEnumerants = True,
                  helper_file_type = ''):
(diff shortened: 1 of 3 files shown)


mark-lunarg:
Man, nothing ever just works anymore, does it.

shannon-lunarg:
Haha. Apparently not.

shannon-lunarg:
All of the checks are now passing

PR #418, opened 2020-07-06
build: Update known-good files for 1.2.146 header

shannon-lunarg:
Changes:
- Updated `known_good.json`
- Added directfb to platform dictionary in `common_codegen.py`
- Generated new source files

Parked until the 146 header is integrated in the Loader.


shannon-lunarg:
The loader known-good has been updated for the 146 header. Vulkan-Tools known-good needs to be updated before the 146 header can be integrated into ValidationLayers.

PR #424, opened 2020-07-15
Make CMake project work better as sub-project

tksuoran:
These changes make Vulkan-Tools work better as cmake subproject from a parent master cmake project.

Use CMake targets for Vulkan loader and headers if found.
Make vulkaninfo generated source path work as CMake subproject.


jeremyk-lunarg:
Closed due to inactivity.

PR #425, opened 2020-07-15
Add support for VK_EXT_fragment_density_map2 to mock ICD

janharaldfredriksen-arm:
The VK_EXT_fragment_density_map2 extension has some limits that are required to be > 0, but without initialization by the ICD they will of course be reported as 0. I only needed to initialize one limit (maxDescriptorSetSubsampledSamplers) for my validation code / tests to work, but initializing all limits to their minimum legal values here.


jeremyk-lunarg:
There as been no action on this PR for a few months now, and it is failing CI.
Are there an plans to update soon?

jeremyk-lunarg:
When you change a generator file like `mock_icd_generator.py`, you have to generate the file and commit the changed generated file as well as the generator.  That error is letting you know that you didn't commit the newly generated file with the generator.

This is done to minimize build times on the user end.

I would suggest rebasing, regenerating, and resubmitting with the generator changes and the changed generator file.

Issue #308, opened 2019-12-13
vkcube with VK_KHR_display fails when acquirable displays are on not-the-first physical devices


jjulianoatnv:
The way that vkcube's support for `VK_KHR_display` is implemented, vkcube fails when no direct-to-display `VkDisplayKHR` can be acquired on the `vkPhysicalDevice` that is sorted first in the list of physical devices, even when additional physical devices do have displays that can be acquired.

This is an app bug. It prevents using vkcube to test some multi-GPU scenarios.

This bug happens because vkcube is "lazy" in how it selects a physical device. It always selects the first physical device, even if this physical device does not have any displays that can be acquired.

```    if (gpu_count > 0) {
        VkPhysicalDevice *physical_devices = malloc(sizeof(VkPhysicalDevice) * gpu_count);
        err = vkEnumeratePhysicalDevices(demo->inst, &gpu_count, physical_devices);
        assert(!err);
        /* For cube demo we just grab the first physical device */
        demo->gpu = physical_devices[0];
```

`demo->gpu` is hard-coded to the first physical device without regard to the presence of an acquirable display on the physical device.

cubanismo:
Note this isn't specific to VK_KHR_display.  It will also fail on NVIDIA Linux drivers with multiple GPUs and an X11 screen on each GPU, since we don't currently support cross-GPU presentation on X11.

I think the right way to solve both is to use vkGetPhysicalDeviceSurfaceSupportKHR().  That will only fix the X11 issue on the very latest NVIDIA drivers, but it should fix the VK_KHR_display issue (assuming additional changes cause vkcube to try to enumerate displays on all GPUs before selecting a GPU for actual rendering) everywhere as well.

achea:
As a workaround on an Optimus system, I'm able to set `VK_ICD_FILENAMES=/path/to/nvidia_icd.json`.  I don't know how Vulkan works, but what I think this does is override the list of available drivers to just one, so that `vkcube` will 'lazily' select the only available ICD, which is nvidia's in my case.

This "fix" has been floating around on various wikis and projects, and also works with `vulkaninfo` (i.e. shows just the one nvidia driver, not both intel and nvidia).  I just want to leave this crumb in case anyone else find this issue when testing `vkcube`.

Tony-LunarG:
[...]

PR #427, opened 2020-07-16
Add option to select phys device to vkcube /  vkcubepp

Tony-LunarG:
Workaround for issue in #308

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -337,6 +337,7 @@ struct demo {
     bool use_staging_buffer;
     bool separate_present_queue;
     bool is_minimized;
+    uint32_t gpu_number;
 
     bool VK_KHR_incremental_present_enabled;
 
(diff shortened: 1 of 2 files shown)


cube/cube.c
```diff
@@ -3178,8 +3179,7 @@ static void demo_init_vk(struct demo *demo) {
         VkPhysicalDevice *physical_devices = malloc(sizeof(VkPhysicalDevice) * gpu_count);
         err = vkEnumeratePhysicalDevices(demo->inst, &gpu_count, physical_devices);
         assert(!err);
-        /* For cube demo we just grab the first physical device */
-        demo->gpu = physical_devices[0];
+        demo->gpu = physical_devices[demo->gpu_number];
```
jeremyk-lunarg, line 3182: This segfaults if the number selected by the user is out of bounds.

It would be cleaner if it either:
1. Crashed more gracefully with an error message.
2. Output a warning to the user and use the first gpu in the list.

cube/cube.cpp
```diff
@@ -1215,8 +1223,7 @@ void Demo::init_vk() {
         std::unique_ptr<vk::PhysicalDevice[]> physical_devices(new vk::PhysicalDevice[gpu_count]);
         result = inst.enumeratePhysicalDevices(&gpu_count, physical_devices.get());
         VERIFY(result == vk::Result::eSuccess);
-        /* For cube demo we just grab the first physical device */
-        gpu = physical_devices[0];
+        gpu = physical_devices[gpu_number];
```
jeremyk-lunarg, line 1226: Same as vkcube.

Tony-LunarG: Good point.  I choose option 2.

PR #430, opened 2020-07-28
vulkaninfo: fix device ext list having bad items

charles-lunarg:
vkconfig exposed an issue in the vulkan-loader which resulted in two different sizes
returned from vkEnumerateDeviceExtensionProperties(). A full fix requires a change to
the loader, but this commit stops the issue from happening in vulkaninfo. The issue is
because the count might be less on the second enumerate call, so resizing the vector
again trims the extraneous elements away.

Change-Id: I808d1fd13868711675aae5c91ea5100ec8cbc316

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -240,6 +240,7 @@ auto GetVectorInit(const char *func_name, F &&f, T init, Ts &&... ts) -> std::ve
         if (err) THROW_VK_ERR(func_name, err);
         results.resize(count, init);
         err = f(ts..., &count, results.data());
+        results.resize(count);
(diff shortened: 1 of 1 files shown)


charles-lunarg:
Vulkan-Loader issue in question: https://github.com/KhronosGroup/Vulkan-Loader/issues/403

charles-lunarg:
`If pPropertyCount is smaller than the number of extensions available, VK_INCOMPLETE will be returned instead of VK_SUCCESS, to indicate that not all the available properties were returned.`
The spec for basically all enumerate functions is that the second call will return VK_INCOMPLETE if the count ever increased, which this code handles.

richard-lunarg:
Works for me then.

PR #440, opened 2020-08-24
Fix some non-msvc compiler warnings

Biswa96:


jeremyk-lunarg:
Passing internal and cloud CI.

Merging.

PR #443, opened 2020-09-07
build: Update known-good files for 1.2.152 header

shannon-lunarg:
Changes:
- Updated `known_good.json`

Running through internal CI currently.


jeremyk-lunarg (approved):
Builds and runs on macOS.

PR #449, opened 2020-09-24
build: Fix linux/windows build in single directory

ncesario-lunarg:
Allows a single worktree to be used by multiple platforms for builds.


Tony-LunarG (approved):
LGTM

PR #454, opened 2020-10-01
vulkaninfo: Add more portability macro guards

charles-lunarg:
Previous commit didn't contain all that were necessary.

Change-Id: I0983beb06d95dd316ece5b7ba52754e4ad180cbd


jeremyk-lunarg:
Builds and runs on macOS.

Fix the clang-format and it should be good to go.

PR #457, opened 2020-10-28
Fix uninitialized variable use in the cube demo

Lahvuun:
I was doing some admittedly questionable stuff with LD_PRELOAD and stumbled upon an odd segfault:
```
(gdb) set environment LD_PRELOAD /home/lahvuun/.local/share/Steam/ubuntu12_64/gameoverlayrenderer.so:/home/lahvuun/git/xcbtoxlib/buildrelease/libxcbtoxlib.so
(gdb) r
Starting program: /usr/bin/vkcube
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7f3b8de in xcb_screen_sizeof (_buffer=_buffer@entry=0x5555555d9ff4) at xproto.c:508
508         for(i=0; i<_aux->allowed_depths_len; i++) {
```

For some reason setting LD_PRELOAD changed the value of `scr` in `demo_init_connection`.
Without LD_PRELOAD:
```
(gdb) b demo_init_connection
Breakpoint 1 at 0x4996: file /usr/src/debug/dev-util/vulkan-tools-1.2.141/Vulkan-Tools-1.2.141/cube/cube.c, line 3656.
(gdb) r
Starting program: /usr/bin/vkcube
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Breakpoint 1, demo_init_connection (demo=0x7fffffffcb00) at /usr/src/debug/dev-util/vulkan-tools-1.2.141/Vulkan-Tools-1.2.141/cube/cube.c:3656
3656    static void demo_init_connection(struct demo *demo) {
(gdb) n
3662        const char *display_envar = getenv("DISPLAY");
(gdb) p scr
$1 = 0
```

With:
```
(gdb) set environment LD_PRELOAD /home/lahvuun/.local/share/Steam/ubuntu12_64/gameoverlayrenderer.so:/home/lahvuun/git/xcbtoxlib/buildrelease/libxcbtoxlib.so
(gdb) b demo_init_connection
Breakpoint 1 at 0x4996: file /usr/src/debug/dev-util/vulkan-tools-1.2.141/Vulkan-Tools-1.2.141/cube/cube.c, line 3656.
(gdb) r
Starting program: /usr/bin/vkcube
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Breakpoint 1, demo_init_connection (demo=0x7fffffffca70) at /usr/src/debug/dev-util/vulkan-tools-1.2.141/Vulkan-Tools-1.2.141/cube/cube.c:3656
3656    static void demo_init_connection(struct demo *demo) {
(gdb) n
3662        const char *display_envar = getenv("DISPLAY");
(gdb) p scr
$1 = 32767
```

Which then breaks the `while` check:
```while (scr-- > 0)```
and eventually leads to a segfault while iterating over `xcb_screen`s.

I wonder why the compiler doesn't pick this up?

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -3766,7 +3766,7 @@ static void demo_init_connection(struct demo *demo) {
 #if defined(VK_USE_PLATFORM_XCB_KHR)
     const xcb_setup_t *setup;
     xcb_screen_iterator_t iter;
-    int scr;
+    int scr = 0;
 
     const char *display_envar = getenv("DISPLAY");
     if (display_envar == NULL || display_envar[0] == '\0') {


Tony-LunarG:
Shouldn't ```demo->connection = xcb_connect(NULL, &scr);``` initialize it?  And if not, shouldn't ```if (xcb_connection_has_error(demo->connection) > 0)``` catch it?

Lahvuun:
I looked into it a little bit more and it _does_ get initialized to 0, actually.

The documentation is a bit confusing:
> Connects to the X server specified by displayname. If displayname is NULL, uses the value of the DISPLAY environment variable. If a particular screen on that server is preferred, the int pointed to by screenp (if not NULL) will be set to that screen; otherwise the screen will be set to 0.

What actually happens in the library is that when `displayname` is set to `NULL` but `screenp` is not, it ignores the value and sets it to the screen token at the end of `DISPLAY`. Or, if that's missing, to 0.

I guess in a way it makes sense to discard the value of `screenp` if `displayname` is `NULL`, something my replacement of `xcb_connect` didn't do.

Sorry for the confusion!

PR #458, opened 2020-10-29
build: Update known-good files for 1.2.158 header

shannon-lunarg:
Changes:
- Updated `known_good.json`
- Generated new source files

Passes internal CI


jeremyk-lunarg (approved):
Builds and runs on macOS.

PR #464, opened 2020-11-17
vulkaninfo: add option to always open details for HTML

charles-lunarg:
Doesn't alter current behavior.

Change-Id: I76cb0c4395741848667b2687d722b9b2245fb296

### DIFF
--- a/vulkaninfo/outputprinter.h
+++ b/vulkaninfo/outputprinter.h
@@ -257,6 +257,11 @@ class Printer {
         return *this;
     }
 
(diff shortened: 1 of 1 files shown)


richard-lunarg (approved):
Looks good. This is part of a WIP to support vulkaninfo on iOS. There will be a conditional compile then that always sets should_always_open to true in that environment, as expanding all the little tree items is overly tedious on a small touch device.

jeremyk-lunarg (approved):

PR #465, opened 2020-11-18
build: Update known-good files for 1.2.161 header

shannon-lunarg:
Changes:
- Updated `known_good.json`

Passes internal CI


jeremyk-lunarg (approved):
Builds and runs on macOS.

PR #466, opened 2020-11-18
vulkaninfo: WIP: Modified for use from iOS

richard-lunarg:
Work in progress. Modifications to support use from iOS where console output is not available, but json and portability output is directly captured to .json files. For iOS, main() is now vulkainInfoMain() and is called from an iOS host GUI process still in development.

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -865,14 +865,7 @@ int main(int argc, char **argv) {
 // On iOS, we'll call this ourselves from a parent routine in the GUI
 int vulkanInfoMain(int argc, char **argv) {
 #endif
-    
-    bool human_readable_output = true;
-    bool html_output = false;
-    bool json_output = false;
-    bool vkconfig_output = false;
-    bool portability_json = false;
-    bool summary = false;
-    
+        
 #ifdef _WIN32
     if (ConsoleIsExclusive()) ConsoleEnlarge();
     if (!LoadUser32Dll()) {
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -859,11 +859,11 @@ void print_usage(const char *argv0) {
     std::cout << "--summary           Show a summary of the instance and GPU's on a system.\n\n";
 }
 
-#ifndef VK_USE_PLATFORM_IOS_MVK
-int main(int argc, char **argv) {
-#else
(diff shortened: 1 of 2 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -988,8 +1002,14 @@ int main(int argc, char **argv) {
                     "'s portability features and properties. Generated using the vulkaninfo program.\",\n\t\t\"vulkanApiVersion\": "
                     "\"" +
                     VkVersionString(instance.vk_version) + "\"\n" + "\t}";
+#ifndef VK_USE_PLATFORM_IOS_MVK
                 printers.push_back(
                     std::unique_ptr<Printer>(new Printer(OutputType::json, out, selected_gpu, instance.vk_version, start_string)));
+#else
+                portability_out = std::ofstream("portabiliyt.json");
```
charles-lunarg, line 1009: typo:
"portabiliyt" -> "portability"

vulkaninfo/vulkaninfo.cpp
```diff
@@ -859,7 +859,13 @@ void print_usage(const char *argv0) {
     std::cout << "--summary           Show a summary of the instance and GPU's on a system.\n\n";
 }
 
+#ifndef VK_USE_PLATFORM_IOS_MVK
```
charles-lunarg, line 862: Nitpick, but I find it nicer to not use double negatives.
Here, there would be `#if defined(IOS_MVK)` vulkanInfoMain(...) #else main(...)`

vulkaninfo/vulkaninfo.h
```diff
@@ -108,14 +108,14 @@ struct VulkanException : std::runtime_error {
 };
 #define THROW_VK_ERR(func_name, err) throw VulkanException(func_name, __FILE__, __LINE__, err);
 
-// global configuration
```
charles-lunarg, line 111: Was this an accidental removal? 
Also, it would be nice if it was documented *why* these bools are here instead of in main(). Specifically, that the window's WAIT_FOR_CONSOLE_DESTROY macro needs to know if the program is printing to a console (eg human readible output).

charles-lunarg (changes requested):
Small typo + 2 nitpicks.

vulkaninfo/vulkaninfo.h
richard-lunarg: All good points, and cleaned up.

charles-lunarg (approved):

PR #467, opened 2020-11-20
scripts: Remove assert for nested array types

mikes-lunarg:
Multi-dimensional arrays are already in the specification and don't cause any issues for vulkaninfo


charles-lunarg (approved):
This is an artifact from when I was first writing the autogen, and wanted to be explicit about not supporting multidimentional arrays. Now that they are in the spec ( VkTransformMatrixKHR ) this breaks the autogen but that struct isn't used by vulkaninfo.
Removing the assert fixes the codegen.

PR #468, opened 2020-11-24
build: Update known-good files for 1.2.162 header

shannon-lunarg:
Changes:
- Updated `known_good.json`
- Generated new source files

Running through internal CI


jeremyk-lunarg (approved):
Builds on macOS.

shannon-lunarg:
Passes internal CI 

Do the changes in `b4ef19a` seem reasonable? @jeremyk-lunarg

jeremyk-lunarg:
> Passes internal CI
> 
> Do the changes in `b4ef19a` seem reasonable? @jeremyk-lunarg

A quick search of the spec does not show any reference to `vkBindAccelerationStructureMemoryKHR`.

If it's not in the spec for this header, it makes sense to remove it to me.

shannon-lunarg:
@jeremyk-lunarg I made a much larger change in the second commit than was there previously. @Tony-LunarG pointed out, the proper way to update `vulkan_wrapper.cpp` is by updating `generate_vulkan_wrapper.py` so I went through the process found here: https://github.com/KhronosGroup/Vulkan-Tools/tree/master/common 
I think it would be a good idea to test again on mac. Hopefully this Just Works (TM).

shannon-lunarg:
Passed internal CI again.

jeremyk-lunarg (approved):
Builds on macOS.

shannon-lunarg:
All changes pushed to master.

PR #473, opened 2020-12-09
vulkaninfo: Fix win32 fullscreen pnext chain

charles-lunarg:
vulkaninfo was triggering
VUID-vkGetPhysicalDeviceSurfaceCapabilities2KHR-pNext-02671 by not adding a
`VkSurfaceFullScreenExclusiveWin32InfoEXT` in the pNext chain of
`VkSurfaceCapabilitiesFullScreenExclusiveEXT`. This commit fixes it by adding
the struct and filling in `HMONITOR` with the 'current monitor' as gotten by
`MonitorFromWindow()` with a default of using the primary monitor.

Change-Id: I025f66ef8e2d47ebb29ccd7361d3d162c644de09

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -1219,9 +1219,8 @@ class AppSurface {
                                                                surface_extension.surface};
 
         if (inst.CheckExtensionEnabled(VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME)) {
-            VkSurfaceFormat2KHR init;
+            VkSurfaceFormat2KHR init{};
             init.sType = VK_STRUCTURE_TYPE_SURFACE_FORMAT_2_KHR;
-            init.pNext = nullptr;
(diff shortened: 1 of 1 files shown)


pdaniell-nv (approved):

charles-lunarg:
I have cherry picked this over into the 1.2.162 SDK branch. 
Whether it is in the SDK will depend on whether the build is before or after I cherry-picked it, but its likely to have it.

Can I ask how you spotted this? I looked into the vulkaninfo history and found that this issue has been present for quite a while.

pdaniell-nv:
> Can I ask how you spotted this? I looked into the vulkaninfo history and found that this issue has been present for quite a while.

I just happened to notice it when testing the new vulkaninfo. I noticed that it reported "false" for us, but we've supported full screen for a while. I suspected this issue has been there for a long while, and I only noticed it now.

charles-lunarg:
That makes sense. I checked the change and the un-fixed version by running against the validation layers, and it seems they do not report it, so this seems like a hole in the validation layers for this extension.

PR #477, opened 2020-12-22
Vulkaninfo for iOS

richard-lunarg:
This project creates a Vulkaninfo app for iOS devices. Three files are output that can be obtained via the iOS file sharing mechanism for vulkaninfo:
vulkaninfo.html -> HTML output, which is also displayed on the device when run.
vulkaninfo.json -> Standard json output from vulkaninfo
portability.json -> The portability schema output for the device.

VulkanInfo was also changed slightly as the identified GPU number is now also accompanied by the device name.
iOS specific code is guarded by the #define VK_USE_PLATFORM_IOS_MVK
VK_ENABLE_BETA_EXTENSIONS is also required as this uses the portability extension.


charles-lunarg (approved):
Changes to vulkaninfo/vulkaninfo.cpp looks reasonable and are a nice addition to the json output.

PR #479, opened 2021-01-06
vkcube: Updated to support portability extension properly

richard-lunarg:
Updated vkcube to properly enable the VK_KHR_portability_subset. Once added, the swizzle formats were then tagged as failing validation and had to be changed to VK_COMPONENT_SWIZZLE_IDENTITY, which does not result in a change in behavior.


ncesario-lunarg (approved):
LGTM!

PR #480, opened 2021-01-07
vkcube: Improve GPU auto-selection

baryluk:
Instead of selecting GPU 0, prefer dGPU, iGPU, vGPU, CPU, OTHER.
In that order.

If user selects specific gpu_number (including 0), use only
that and don't fallback to anything else, even when incorrect.

### DIFF
--- a/cube/cube.cpp
+++ b/cube/cube.cpp
@@ -311,7 +311,7 @@ struct Demo {
     bool use_staging_buffer;
     bool use_xlib;
     bool separate_present_queue;
-    uint32_t gpu_number;
+    int32_t gpu_number;
 
     vk::Instance inst;
     vk::PhysicalDevice gpu;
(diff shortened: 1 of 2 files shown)


baryluk:
Ready for review.

cube/cube.cpp
```diff
@@ -1243,29 +1244,67 @@ void Demo::init_vk() {
...
+                gpu_number = i;
+                break;
+            }
+        }
+    }
+    assert(gpu_number >= 0);
+    gpu = physical_devices[gpu_number];
+    {
+        auto physicalDeviceProperties = gpu.getProperties();
+        fprintf(stderr, "Selected GPU %d: %s, type: %s\n", gpu_number, physicalDeviceProperties.deviceName.data(),
```
Tony-LunarG, line 1303: This isn't really an error.  I'd rather not change the output of cube (or cubepp) as we are aware of people using them in testing and changes in output could upset results.

Tony-LunarG (changes requested):
LGTM if no output changes

cube/cube.cpp
baryluk: Could you give a more specific example of people using them in testing?

baryluk: I believe more clean device selection diagnostic, for vkcube, would make it much more useful to humans. Inconvenience of breaking maybe one test script (if any), is worth it IMHO, and anybody who relays on some very exact output of a graphical tool used mostly by humans, can blame themselves.

Tony-LunarG: Ok, fair enough

Tony-LunarG (approved):

PR #481, opened 2021-01-08
vkcube: Add options --width and --height to specify width and height

caramelli:
This change adds --w and --h options to specify width and height when running vkcube.

Nicolas Caramelli


Tony-LunarG (changes requested):
Looks good, just one ask (and rebase).  Could you go with --height and --width or -h and -w, which seems to be the modern convention.  Yes, there's --c, but I'd rather not perpetuate that style.

caramelli:
OK, the --w and --h were proposed for consistency with --c but I agree, it's better to have two dashes with long options. 

Nicolas Caramelli

Tony-LunarG:
Rebased and pushed to master.  Thanks!

PR #495, opened 2021-03-24
build: Update known-good files for 1.2.173 header

mikes-lunarg:
- Update known-good files
- Generate source code


jeremyk-lunarg (approved):
Builds and runs on macOS.
LGTM

mikes-lunarg:
Passed internal CI

PR #497, opened 2021-03-29
build: Update to header 1.2.174

mikes-lunarg:
- Update known-good
- Generate source (no changes)

Change-Id: Idb543c5c0b078db219c6cf60dc03f4d46460e808


jeremyk-lunarg (approved):
Builds on macOS.
LGTM

PR #502, opened 2021-04-20
build: Update to header 1.2.176

mikes-lunarg:
- Update known-good
- Generate source
- Fix scripts/generate_vulkan_wrapper.py
  - Handle new version macros in vulkan_core.h
  - Recognize beta/provisional platform guard
  - Skip extensions without commands


common/vulkan_wrapper.cpp
```diff
@@ -1048,6 +1085,38 @@ PFN_vkCreateScreenSurfaceQNX vkCreateScreenSurfaceQNX;
...
+PFN_vkCmdEndVideoCodingKHR vkCmdEndVideoCodingKHR;
+PFN_vkCmdControlVideoCodingKHR vkCmdControlVideoCodingKHR;
+#endif
+
+#ifdef VK_USE_PLATFORM_SCREEN_QNX
+PFN_vkCmdDecodeVideoKHR vkCmdDecodeVideoKHR;
+#endif
+
+#ifdef VK_USE_PLATFORM_SCREEN_QNX
+#endif
```
jeremyk-lunarg, line 1108: What's with the empty `ifdef`?

icd/generated/mock_icd.cpp
```diff
@@ -3568,6 +3698,12 @@ static VKAPI_ATTR void VKAPI_CALL CmdDrawIndexedIndirectCountAMD(
 
 
 
+#ifdef VK_ENABLE_BETA_EXTENSIONS
+#endif /* VK_ENABLE_BETA_EXTENSIONS */
+
+#ifdef VK_ENABLE_BETA_EXTENSIONS
+#endif /* VK_ENABLE_BETA_EXTENSIONS */
+
```
jeremyk-lunarg, line 3706: Looks like the empty `ifdef`s are throughout the code.

jeremyk-lunarg (approved):
A few nitpicks, but LGTM.
Just wait for Actions to finish up.
Ignore the AppVeyor failure.  We're moving away from that, but it's still running for some reason.

icd/generated/mock_icd.cpp
mikes-lunarg: The video extension is actually multiple extensions, many of which have no entry points. So the code generator will open/close the platform block on `beginFeature()`/`endFeature()` callbacks, but never receive any callbacks or emit any code in between them. This is not unique to `VK_ENABLE_BETA_EXTENSIONS`, you can also find this pattern for some of the `VK_USE_PLATFORM_WIN32` and `VK_USE_PLATFORM_GGP` ifdefs

common/vulkan_wrapper.cpp
mikes-lunarg: I'm not too concerned about the empty `#ifdef`s, but looking closer, the video content should be wrapped with `VK_ENABLE_BETA_EXTENSIONS` instead of `VK_USE_PLATFORM_SCREEN_QNX`. I think generate_vulkan_wrapper.py isn't detecting the correct platform guard for beta content.

icd/generated/mock_icd.cpp
jeremyk-lunarg: Ah, okay.

common/vulkan_wrapper.cpp
mikes-lunarg: Pushed a fix in debeff612dfaaa338e0d705423501a5ad63c05e5 to for the beta platform guards and it was easy enough strip out empty `#ifdef` blocks at the same time.

PR #513, opened 2021-05-17
vulkaninfo: Add check for printing memory budget

jeremyk-lunarg:
Added check for the `VK_EXT_memory_budget` before printing memory usage
and memory budget under VkPhysicalDeviceMemoryProperties.


charles-lunarg (approved):
Looks good to me.
Better to print nothing than to print junk.

PR #515, opened 2021-05-18
vulkaninfo refactor

charles-lunarg:
Refactor of vulkaninfo

Manually diffed all 4 outputs and there were no changes to the output (save for memory usage which isn't static even across runs).

Changes include:
* Update License Year
* Simplify QueueFamilyProperties by not querying `VkQueueFamilyProperties2` and making the AppQueueFamilyProperties constructed inside the AppGpu object
* Make printers use the original struct where before they would branch on using the new or old one. (props vs props2 for instance)
* Add helper function to create `VulkanVersion` struct
* Add `optional<T>` type for increased legibility.
* Remove consecutive platform guards
* Initialize variables where appropriate
* Remove unused variables
* Remove unused functions
* Make use of standard algorithms instead of raw for loops.
* Reorder arguments of `GetImageCreateInfo` for consistency

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -545,6 +545,43 @@ void freepNextChain(VkStructureHeader *first) {
     }
 }
 
+/* An ptional contains either a value or nothing. The trivial_optional asserts if a value is trying to be gotten but none exist.
+ * The interface is taken from C++17's <optional> with many aspects removed.
+ * This class assumes the template type is 'trivial'
+ */
+namespace util {
+template <typename T>
+struct trivial_optional {
+    using value_type = T;
+
+    bool _contains_value = false;
+    T _value;
+
+    void check_type_properties() {
+        static_assert(std::is_trivially_constructible<T>::value, "Type must be trivially constructable");
+        static_assert(std::is_trivially_copyable<T>::value, "Type must be trivially copyable");
+    }
+
+    trivial_optional() noexcept : _contains_value(false), _value({}) { check_type_properties(); }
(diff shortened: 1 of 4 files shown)


jeremyk-lunarg:
This repository advertises support for Visual Studio 2013, 2015, and 2017.

The `std::optional<T>` template was introduced in C++17.  Will this change affect backwards compatibility?

vulkaninfo/vulkaninfo.h
```diff
@@ -545,8 +525,45 @@ void freepNextChain(VkStructureHeader *first) {
     }
 }
 
+/* An ptional contains either a value or nothing. The optional asserts if a value is trying to be gotten but none exist.
+ * The interface is taken from C++17's <optional> with many aspects removed.
+ * This type assumes the optional type is 'trivial'
+ */
+namespace util {
+template <typename T>
+struct optional {
```
jeremyk-lunarg, line 534: The implementation itself looks alright, but to avoid confusion with the official STL version of `std::optional<T>` it might be a good idea to rename it to something like `optional_lite<T>`.

jeremyk-lunarg (changes requested):
I don't mind the use of the new `util::optional<T>` datatype, but to avoid confusion in the future I think it should be renamed.

vulkaninfo/vulkaninfo.h
charles-lunarg: I can change the name, though I think `trivial_optional` would likely be more suitable as it only works with trivial types.

charles-lunarg: Fixed.

jeremyk-lunarg (approved):

PR #516, opened 2021-05-23
Trigger demo quit from applicationWillTerminate

Zorro666:
#### Outcome 

A clean shutdown of the rendering thread (`CVDisplayLink`) whilst the main thread is exiting.

#### Problem

Locally I am working with the Mac version of `RenderDoc` and using samples like the cube projects as part of my local manual testing of the `Vulkan` implementation.

Whilst capturing the cube project with `RenderDoc`. I discovered a crash during shutdown which only happens when `RenderDoc` is attached. I tracked this back to the cube project rendering thread bring active whilst the main thread exit flow is executing ie. static destructors. 

#### Notes

This PR is not critical. I am happy to keep it local to help my testing of the Mac `Vulkan` version of `RenderDoc`.

This PR is one proposed way to fix the shutdown crash when capturing from `RenderDoc` by quitting the demo from the application event `applicationWillTerminate`.

I also changed the exiting `dealloc` implementation because it appeared to be calling things in the wrong ie. demo cleanup was called before stopping the rendering thread which in principal could have left the rendering thread active whilst the demo was being deleted. In my local testing the `dealloc` method did not appear to be called.

I have a different (a bit simpler) implementation to solve the problem which is to stop and start the `CVDisplayLink` from the `DemoViewController` events `viewWillDisappear`  & `viewWillAppear`.

#### Testing

Locally on M1 MacBookAir for the vkcube & vkcubepp applications (debug and release builds) attached and unattached to RenderDoc Mac.

### DIFF
--- a/cube/macOS/cube/AppDelegate.m
+++ b/cube/macOS/cube/AppDelegate.m
@@ -17,6 +17,7 @@
  */
 
 #import "AppDelegate.h"
+#import "DemoViewController.h"
 
 @interface AppDelegate ()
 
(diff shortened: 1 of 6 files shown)


richard-lunarg (approved):
Looks good to me. Thanks for cleaning it up in both cube and cubepp.

jeremyk-lunarg:
There was a problem with our Github Actions which should be fixed on `master`.
Please rebase and push again.

PR #521, opened 2021-06-07
build: Update to header 1.2.180

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):
Vulkaninfo changes look good.

jeremyk-lunarg (approved):
LGTM as long as it passes CI.

Issue #469, opened 2020-11-30
Update vkcube to wayland xdg-shell


cdotstout:
In the following commit the change was reverted.  Are there any details of the weston incompatibility or any ticket tracking the steps to reproduce?  It would be nice to move off the deprecated protocol.

https://github.com/KhronosGroup/Vulkan-Tools/commit/5ceb7be9a3bf86f34ade5a7f72459248d99a4b76#diff-4f3c4172a4fe308cebc840da665171f534849b1fca101eaf635d68fb453393db

jbeich:
CC @mstoeckl. Maybe related to the following crash on `weston`, `kwin_wayland`, `wayfire` (wlroots), `hikari` (wlroots), `labwc` (wlroots) despite working fine on `sway` (wlroots), `cage` (wlroots):

```c
$ vkcube-wayland
Selected GPU 0: Intel(R) HD Graphics 530 (SKL GT2), type: 1
=================================================================
==7240==ERROR: AddressSanitizer: heap-use-after-free on address 0x612000156d00 at pc 0x0000002ca714 bp 0x7fffffffc310 sp 0x7fffffffc308
READ of size 8 at 0x612000156d00 thread T0
    #0 0x2ca713 in demo_update_data_buffer Vulkan-Tools-1.2.178/cube/cube.c:894:66
    #1 0x2f5270 in demo_draw Vulkan-Tools-1.2.178/cube/cube.c:1043:5
    #2 0x2d461c in demo_run Vulkan-Tools-1.2.178/cube/cube.c:2757:13
    #3 0x2cec4d in main Vulkan-Tools-1.2.178/cube/cube.c:4267:5

0x612000156d00 is located 192 bytes inside of 288-byte region [0x612000156c40,0x612000156d60)
freed by thread T0 here:
    #0 0x29b2e2 in free (vkcube-wayland+0x29b2e2)
    #1 0x2e08a5 in demo_resize Vulkan-Tools-1.2.178/cube/cube.c:2447:5
    #2 0x2df646 in handle_surface_configure Vulkan-Tools-1.2.178/cube/cube.c:2768:9
    #3 0x8009730dc in ffi_call_unix64 libffi-3.3/src/x86/unix64.S:101

previously allocated by thread T0 here:
    #0 0x29b42d in malloc (vkcube-wayland+0x29b42d)
    #1 0x2e483b in demo_prepare_buffers Vulkan-Tools-1.2.178/cube/cube.c:1363:36
[...]

PR #527, opened 2021-06-11
Rebase and patch "Port Wayland impl from wl-shell to xdg-shell"

Tony-LunarG:
Addresses #469 
@mstoeckl @jbeich please review - I'm counting on your testing for this PR


jbeich:
Works fine under Sway and labwc. I didn't retest other compositors because this PR matches what FreeBSD *already* uses downstream.

mstoeckl:
Same; I tested sway/kwin/gnome/weston and did not encounter any problems.

jeremyk-lunarg (approved):

PR #533, opened 2021-06-15
Rwright vkcube crash fix

richard-lunarg:
On macOS, vkCube is crashing on exit because

CVDisplayLinkRelease(_displayLink)

ends up getting called twice on shutdown. This commit removes one of these calls from the shutdown code path.


jeremyk-lunarg:
Why are there 7 merge commits?

richard-lunarg:
>Why are there 7 merge commits?

I do not know... all of the commits are the same one line changes to the same two files, but with different dates though.

richard-lunarg:
No, actually they aren't. I think this was because my master was behind when I created the branch. I'm going to close this PR and start over with a new branch.

PR #537, opened 2021-06-15
vkcube: fix crash on exit from double delete

richard-lunarg:
Fixed crash on exit for vkcube due to

    CVDisplayLinkRelease(displayLink);

Being called twice on shutdown. Removed redundant call that was added recently.


jeremyk-lunarg (approved):
LGTM as long as it passes CI.

PR #539, opened 2021-06-17
cube: Handle occlusion where currentExtent == 0xFFFFFFFF

Joshua-Ashton:
The Vulkan spec states that: "On some platforms, it is normal that maxImageExtent may become (0, 0), for example when the window is minimized. In such a case, it is not possible to create a swapchain due to the Valid Usage requirements."

Yet vkcube previously only handled this path for platforms where currentExtent != 0xFFFFFFFF.

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -1237,7 +1237,7 @@ static void demo_prepare_buffers(struct demo *demo) {
         demo->height = surfCapabilities.currentExtent.height;
     }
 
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
Changes look good.
> Averts the case when the current extent is `0xFFFFFFFF` and demo->width/height isn't updated but the max_extent is 0 and should have vkcube be in a minimized state and isn't thus causing crashes later on.

PR #540, opened 2021-06-17
cube: Don't constantly call vkAllocateCommandBuffers and vkBeginComma…

Joshua-Ashton:
…ndBuffer when minimised

Before this, vkcube would leak these crash when minimised for long periods of time.


Joshua-Ashton:
For context, it tries to allocate these constantly when minimised and they are never freed as that's behind a `if  (!demo->is_minimized) {` check.

charles-lunarg (approved):
Very silly to have a 'check to see if we should return' _after_ doing creation of resources, so this is a good fix.

PR #541, opened 2021-06-21
vulkaninfo: Allow specifying file output

charles-lunarg:
This PR fixes up the code base so that adding the ability to specify a file to print to is simple. To do this requires:
* Creating a `OutputCategory` enum to hold which type of output `vulkaninfo` should print.
* Refactoring the `user32.dll` code so that it doesn't use any macros and globals are limited to a single pointer to a `User32DllHandle` class (for use in a Win32 callback)
* Creating three functions, `parse_arguments`, `get_printer_create_details`, and `RunPrinter`, that reduce the size of `main()` to a manageable level
* Clean up the logic to decide the destination of the output (stdout or file) and the start string into a very tidy switch statement.

### DIFF
--- a/vulkaninfo/outputprinter.h
+++ b/vulkaninfo/outputprinter.h
@@ -735,9 +735,7 @@ class ObjectWrapper {
 
 class ArrayWrapper {
   public:
-    ArrayWrapper(Printer &p, std::string array_name, size_t element_count = 0) : p(p) {
-        p.ArrayStart(array_name, element_count);
-    }
+    ArrayWrapper(Printer &p, std::string array_name, size_t element_count = 0) : p(p) { p.ArrayStart(array_name, element_count); }
     ~ArrayWrapper() { p.ArrayEnd(); }
 
   private:
--- a/vulkaninfo/outputprinter.h
+++ b/vulkaninfo/outputprinter.h
@@ -65,11 +65,17 @@ std::string VkVersionString(VulkanVersion v) {
 
 enum class OutputType { text, html, json, vkconfig_output };
 
+struct PrinterCreateDetails {
+    OutputType output_type = OutputType::text;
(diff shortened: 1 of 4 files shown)


jeremyk-lunarg (approved):
LGTM

charles-lunarg:
There are a ton of changes to the code made and I'm not confident there weren't some little bugs that crept in. I therefore will need to do some more testing on it before I can feel safe I'm not introducing regressions.

jeremyk-lunarg:
Why did you rename `trivial_optional` to `optional`?
Are we using the `optional` from the C++ standard library?

charles-lunarg:
Good comment.
I altered the implementation to allow non-trivial types. It originally required trivial types because I wanted to keep the custom optional type simple. But it limited usefulness since I couldn't put `std::vector`s or `std::string`s in it. Looking at the implementation, I didn't see a reason it couldn't have non-trivial types, there wasn't any unioning or type erasure going on. 

It does not use the C++17 standard library type, though it models the interface on it. *Ideally* the C++17 implementation would be a drop in replacement, but I don't want to increase the C++ version required any more than necessary.

The rename is to signify that it can use non-trivial types. Better to change the name than be misrepresentative. Open to better alternate names, to avoid confusion with `std::optional`.

jeremyk-lunarg:
I'm not great at coming up with names.
Maybe `util::vulkaninfo_optional`?

charles-lunarg:
Thats a better name than just `optional`, since it won't be confused with the C++17 library name.

jeremyk-lunarg (approved):

charles-lunarg:
One last force push (hopefully), I found that the vkconfig_output had a bug where it required 2 parameters (the old 'dir' parameter) but now doesn't need to do that. Once it passes CI, I will merge it.

PR #546, opened 2021-06-28
vulkaninfo: Fix device version check and building for Android

dreifachstein:


charles-lunarg:
Vulkaninfo is not currently supported on android, and this PR is far from capable of supporting it. Beyond the CI failures due to lack of platform guards in the CMakeLists.txt, the build script uses a hardcoded path to a file on the users system. This is not acceptable for a build script that should work on any platform. 

Because of that, I am going to close this PR. Whlie I am hesitant to add Android, its not out of the realm of possibility, but a lot more effort would need to go into a PR before I would accept it. 

The fix for the props version in AppGpu is a real bugfix, so I would appreciate a PR for that issue alone.
Lastly, I am very unsure what the changes to the autogen are for. They seem completely superfluous.

dreifachstein:
Sorry, a local change got mixed in and has been fixed.

PR #547, opened 2021-06-29
windows: updated runtime copyrights to 2017, 2021

lunarpapillo:
This is needed to upright the copyright notice for the Vulkan SDK Runtime Installer, as per:

https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/-/issues/760

I'm not sure whether all the copyrights should be updated,
or just the ones at the top (as the others haven't been modified
since 2016, apparently).  Hopefully reviewers will tell me
whether this is correct.

I also corrected the spelling of "open-source".

### DIFF
--- a/windows-runtime-installer/VulkanRT-License.txt
+++ b/windows-runtime-installer/VulkanRT-License.txt
@@ -1,8 +1,8 @@
-Copyright (c) 2015-2020 The Khronos Group Inc.
-Copyright (c) 2015-2020 LunarG, Inc.
-Copyright (c) 2015-2020 Valve Corporation
+Copyright (c) 2015-2021 The Khronos Group Inc.
+Copyright (c) 2015-2021 LunarG, Inc.
+Copyright (c) 2015-2021 Valve Corporation
 
-The Vulkan Runtime is comprised of 100% open source components (MIT, and
+The Vulkan Runtime is comprised of 100% open-source components (MIT, and
 Apache 2.0). The text of such licenses is included below along with the
 copyrights.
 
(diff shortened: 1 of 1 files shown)


KarenGhavam-lunarG:
It is my understanding we should only update copyrights on files that were edited that year.

lunarpapillo:
> It is my understanding we should only update copyrights on files that were edited that year.

This is the overall copyright for the Runtime Installer, which seems to include a general license, an Apache 2.0 license, and two separate MIT licenses.  I can't tell from the file which license applies to what - is there another way I can figure this out?

My assumption has been that since the Loader itself has been updated in the last year, and the loader is pretty much the core component of the Runtime, that all copyrights would have to change.

But if one of the license clauses actually applies to `vulkaninfo` or something else shipped in the runtime, then we wouldn't generally want to update it... but I can't tell for certain.  And we have updated `vulkaninfo` since 2016, so even then the copyright date is currently incorrect...

lunarpapillo:
Karen pointed me to the LunarG Vulkan License Registry, which neatly solves a lot of these problems.

A quick check shows that `libvulkan` has only four files with an MIT license:
- `cJSON.c` (Dave Gamble)
  - updated 20180417 `tobine@google.com`
  - updated 20171026 `mark@lunarg.com`
- `cJSON.h` (Dave Gamble)
  - updated 20170126 `markl@lunarg.com1
- `murmurhash.c` (joseph werle)
  - updated 20180417 `tobine@google.com`
  - updated 20170126 `mark@lunarg.com`
- `murmurhash.h` (joseph werle)
  - updated 20160209 `karl@lunarg.com`

So the Runtime should probably have a Google copyright too, but that's not my job and I cannot legally attribute it.  But the "Dave Gamble" section of the license should IMHO have its Khronos, Valve, and LunarG copyright updated to 2017 (as this represents the last time LunarG changed the file), and the "joseph werle" section of the license should also be updated to 2017 for the same reasons.

*Further*, the **Vulkan-Loader** files `loader/cJSON.c`, `loader/cjson.h`, and `loader/murmurhash.c` should themselves be updated with a correct copyright header as a separate PR, to ensure that the copyright registry is as accurate as possible.

Does that make sense to everyone?

lunarpapillo:
The corresponding **Vulkan-Loader** change is here: https://github.com/KhronosGroup/Vulkan-Loader/pull/611

jeremyk-lunarg (approved):
If it's good with Karen, it's good with me.

PR #551, opened 2021-07-03
vulkaninfo/cube: added support for FreeBSD

hikiko:
Made some changes for Vulkan-Tools to compile on FreeBSD.


Tony-LunarG (approved):
Cube changes are fine by me

PR #552, opened 2021-07-06
build: Update to header 1.2.184

mikes-lunarg:
- Update known-good
- Generate source

Change-Id: I6f8219e31038b0efe1e141147292d041655dd640


charles-lunarg (approved):
Vulkaninfo changes look fine. At first I was confused because the RDMA_CAPABLE_BIT was present but the features struct wasn't, then I remembered that vulkaninfo doesn't generate features structs for non KHR/EXT extensions.

jeremyk-lunarg (approved):

PR #554, opened 2021-07-21
Add missing generated files and CI check

mikes-lunarg:
I missed a couple files in 1.2.185 header update, and the CI check that would have detected it is also missing. This PR fixes both.


mikes-lunarg:
CI is clean with the new consistency check, going to go ahead and merge this

PR #555, opened 2021-07-27
build: Update to header 1.2.186

mikes-lunarg:
- Update known-good
- Generate source (no change)

Change-Id: I529e0a7175f18f50ae717c4caae27ef066b80839


jeremyk-lunarg (approved):
LGTM as long as it passes CI.

PR #558, opened 2021-08-17
build: Update to header 1.2.189

mikes-lunarg:
- Update known-good
- Generate source (no change)


jeremyk-lunarg (approved):

jeremyk-lunarg (commented):
It seems to be failing CI

mikes-lunarg:
> It seems to be failing CI

Looks like it was github issue. I re-ran the jobs and they are passing now.

PR #562, opened 2021-08-30
build: Update to header 1.2.190

mikes-lunarg:
- Update known-good
- Generate source


jeremyk-lunarg (approved):
LGTM as long as it passes CI.

charles-lunarg (approved):
Code gen looks to be all in the right place

PR #564, opened 2021-09-04
vulkaninfo: Fix error message when gpus.size()==0

CyberShadow:
Avoid an amusing and slightly misleading error message:

"The selected gpu (0) is not a valid GPU index. The available GPUs are in the range of 0 to 18446744073709551615."


Tony-LunarG (approved):
LGTM

PR #568, opened 2021-09-17
Vulkaninfo update autogen for 64 bit flags

charles-lunarg:
Vulkaninfo's current autogen breaks on 64 bit flags/bitmasks. While they are not currently necessary for generating the necessary autogen, they may be in the future and it would be wise to update the repo to support them.

The changes include making use of f-strings everywhere. I verified the changes to the codegen as since it generates the same file each time, and they output is identical, there are no breaking changes. Hopefully the use of f-strings make the code-gen more legible.


mikes-lunarg (approved):
LGTM, of course the proof will be once someone creates a feature/property struct containing a 64-bit enum. I love the f-strings!

jeremyk-lunarg (approved):
LGTM

PR #571, opened 2021-09-28
build: Update update_deps.py

charles-lunarg:
CMake really prefers forward slashes for everything. This commit replaces
the backslashes found from windows with forward slashes. Also updates Update_deps.py
to match VVL.


mikes-lunarg:
This might be a good opportunity to sync the entire script (if it works) because it looks like Loader and ValidationLayers have gotten ahead of the copy in this repo

charles-lunarg:
Good idea, I can do that real quick.

jeremyk-lunarg (approved):
LGTM

PR #572, opened 2021-09-28
vulkaninfo: Generate vulkaninfo.rc file

charles-lunarg:
Previously whenever a SDK was being built, it would have to manually find and
replace the version data in vulkaninfo.rc file with the appropriate values. This 
commit makes it simpler by providing a CMake build argument which allows 
setting the version to use from the command line.


vulkaninfo/CMakeLists.txt
```diff
@@ -18,7 +18,29 @@
 # CMakeLists.txt file for building Vulkaninfo
 
 if(WIN32)
-    add_executable(vulkaninfo vulkaninfo.cpp vulkaninfo.rc)
+    # ~~~
+    # Setup the vulkaninfo.rc flie to contain the correct info
```
jeremyk-lunarg, line 22: Nit: "file" is misspelled.

jeremyk-lunarg (approved):
LGTM with 1 nitpick.

vulkaninfo/CMakeLists.txt
charles-lunarg: Nice catch!

PR #574, opened 2021-10-02
vkcube: Reorder setting of pipeline stage flags

hikiko:
Pipeline stage flags should be set before assigned to pWaitDstStageMask
in VkSubmitInfo, otherwise the assigned pointer is invalid.


Tony-LunarG (approved):
Ok

PR #586, opened 2021-11-15
vulkaninfo: Prevent drivers writing out of bounds

### LINKED ISSUE #580: vulkaninfo: Heap corruption in freepNextChain
Running vulkaninfo on my Windows 11 laptop, Geforce RTX 3060 (462.21)and Intel graphics results in HEAP CORRUPTION in the second call to  freepNextChain(static_cast<VkStructureHeader *>(features2.pNext));  Bisected to 1.2.190 header update.  It happens freeing a VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_INTEGER_DOT_PRODUCT_FEATURES_KHR.

charles-lunarg:
The structure VkPhysicalDeviceShaderIntegerDotProdcutFeaturesKHR contains only a single
feature boolean flag. However, before the final version was published to the public, it
had many boolean flag values. Pre-release drivers often contain support for unpublished
extensions, such as VK_KHR_shader_integer_dot_product. Because the final version was
much smaller than the pre-release version, several drivers try to write to members that
do not exist in the final version, which cause out of bounds writes to the features struct
that vulkaninfo passes into the driver.

By increasing the size of the features struct allocation manually, we can prevent potential
crashes from OOBs writes from drivers.

Fixes #580 

Only discussion item I have is how much documentation of which driver versions caused the crashes should be in the code itself.

### DIFF
--- a/scripts/vulkaninfo_generator.py
+++ b/scripts/vulkaninfo_generator.py
@@ -605,7 +605,14 @@ def PrintChainBuilders(listName, structures, all_structures):
         if s.name in structures:
             out += AddGuardHeader(s)
             if s.sTypeName is not None:
-                out += f"        {{{s.sTypeName}, sizeof({s.name})}},\n"
+                out += f"        {{{s.sTypeName}, sizeof({s.name})"
+                # Specific versions of drivers have an incorrect definition of the size of this struct.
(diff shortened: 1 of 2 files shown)


ziga-lunarg (approved):
LGTM, fixes my crash on intel

PR #597, opened 2021-11-30
vkcube/vkcubepp: Add option to force validation errors

MarkY-LunarG:
During tests, it's hard to determine if the --validation option is
actually working properly.  So add an option (--force_errors) to
intentionally force a few validation error messages to make sure
that validation is working properly.

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -446,6 +446,7 @@ struct demo {
     bool validate_checks_disabled;
     bool use_break;
     bool suppress_popups;
+    bool force_errors;
 
     PFN_vkCreateDebugUtilsMessengerEXT CreateDebugUtilsMessengerEXT;
     PFN_vkDestroyDebugUtilsMessengerEXT DestroyDebugUtilsMessengerEXT;
@@ -633,6 +634,10 @@ static void demo_flush_init_cmd(struct demo *demo) {
 
     VkFence fence;
     VkFenceCreateInfo fence_ci = {.sType = VK_STRUCTURE_TYPE_FENCE_CREATE_INFO, .pNext = NULL, .flags = 0};
+    if (demo->force_errors) {
+        // Remove sType to intentionally force validation layer errors.
+        fence_ci.sType = 0;
+    }
     err = vkCreateFence(demo->device, &fence_ci, NULL, &fence);
     assert(!err);
 
@@ -1445,6 +1450,12 @@ static void demo_prepare_depth(struct demo *demo) {
         .flags = 0,
         .viewType = VK_IMAGE_VIEW_TYPE_2D,
     };
+
+
+    if (demo->force_errors) {
+        // Intentionally force a bad pNext value to generate a validation layer error
+        view.pNext = &image;
+    }
 
     VkMemoryRequirements mem_reqs;
     VkResult U_ASSERT_ONLY err;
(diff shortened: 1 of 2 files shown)


cube/cube.c
```diff
@@ -1446,6 +1451,12 @@ static void demo_prepare_depth(struct demo *demo) {
         .viewType = VK_IMAGE_VIEW_TYPE_2D,
     };
 
+
+    if (demo->force_errors) {
+        // Intentionally force a bad pNext value to generate a validation layer error
+        view.pNext = &image;
```
charles-lunarg, line 1457: Will this cause a crash? bad pnext values are not able to be caught by validation layers since validation doesn't have a clue whether the pointer is valid or not until it de references it, at which point it should crash. Its one thing to force a validation error to be emitted, another to crash outright. If thats intentional, I'm fine with it, just wanting to be clear.

MarkY-LunarG: Nope, works fine for both.

charles-lunarg: Is that because at this point, image is null? I see that vkCreateImage is below this line here.

MarkY-LunarG: No, it's not allowed in the pNext chain so it ignores it.
```
Selected GPU 0: NVIDIA GeForce GTX 1650, type: DiscreteGpu
VUID-VkImageViewCreateInfo-pNext-pNext(ERROR / SPEC): msgNum: 151814321 - Validation Error: [ VUID-VkImageViewCreateInfo-pNext-pNext ] Object 0: handle = 0x283dba8, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x90c80b1 | vkCreateImageView: pCreateInfo->pNext chain includes a structure with unexpected VkStructureType VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO; Allowed structures are [VkImageViewASTCDecodeModeEXT, VkImageViewUsageCreateInfo, VkSamplerYcbcrConversionInfo, VkVideoProfileKHR, VkVideoProfilesKHR]. This error is based on the Valid Usage documentation for version 189 of the Vulkan header.  It is possible that you are using a struct from a private extension or an extension that was added to a later version of the Vulkan header, in which case the use of pCreateInfo->pNext is undefined and may not work correctly with validation enabled The Vulkan spec states: Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid instance of VkImageViewASTCDecodeModeEXT, VkImageViewUsageCreateInfo, VkSamplerYcbcrConversionInfo, VkVideoProfileKHR, or VkVideoProfilesKHR (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-VkImageViewCreateInfo-pNext-pNext)
    Objects: 1
        [0] 0x283dba8, type: 3, name: NULL
```

charles-lunarg: Ahhhh
I saw `view.pNext = &image;` and assume it was a pointer to a `VkImage`, not `VkImageCreateInfo`. Thus it makes sense why it catches the error and doesn't crash. Apologies for the misunderstanding.

charles-lunarg (approved):
Changes look good to me. Would be good to update the SDK test documentation to mention the new command line argument where appropriate.

PR #598, opened 2021-12-01
build: Update to header 1.2.201

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):

charles-lunarg:
Allowing the merge to happen even though the github-actions says it 'failed'. This is because the windows-2016 runner was taken offline today for maintenance in preparation for its removal in march 2022. More details found here.
https://github.blog/changelog/2021-10-19-github-actions-the-windows-2016-runner-image-will-be-removed-from-github-hosted-runners-on-march-15-2022/

Issue #590, opened 2021-11-21
vkcube always uses the X11 WSI.


kkartaltepe:
vkcube is configured such that it will only run on X11 based surfaces even when running under wayland as packaged by all linux distributions. As its a standard test utility highly available it can often lead to misrepresenting vulkan as being available on wayland compositors based on xwayland capabilities.

It would be preferable if vkcube could default to wayland if available and/or allow users to choose the surface they wish to use with flags (such that vkcube would fail to initialize if vulkan couldnt load under wayland for whatever reason).

You can see that the WSI is chosen at compile time and will be XCB by default (and no linux distribution changes this default) https://github.com/KhronosGroup/Vulkan-Tools/blob/master/cube/CMakeLists.txt#L93 . This results in the hardcoding of the WSI via `VK_USE_PLATFORM_XCB_KHR`. 

Another option would be to consider updating cmake to build an additional wayland target such as `vkcube_wayland` similar to the mesa demo's utilities `eglgears_{x11,wayland}` or the existing `vkcubepp` additional target.

jbeich:
Duplicate of #307?

> only run on X11 based surfaces even when running under wayland as packaged by all linux distributions

DragonFly and FreeBSD do have `vkcube-wayland` and `vkcube-display` since https://github.com/freebsd/freebsd-ports/commit/acc8ee9744f6.

> ` eglgears_{x11,wayland}`

`eglgears_wayland` is rotten, see [mesa-demos#16](https://gitlab.freedesktop.org/mesa/demos/-/issues/16).

kkartaltepe:
> eglgears_wayland is rotten, see mesa-demos#16.

Yes I also reached out to mesa-demos and they started to take a look at https://gitlab.freedesktop.org/mesa/demos/-/merge_requests/32 again which updates `eglgears_wayland` for modern times, which mostly works and seems discussion is revived.

> DragonFly and FreeBSD do have vkcube-wayland and vkcube-display since freebsd/freebsd-ports@acc8ee9.

It seems the maintainers for BSD are especially attentive. I doubt any other maintainers understand how vkcube cmake works and as such it would be best for vulkan-tools to provide the build targets instead of depending on maintainers to understand the repository so deeply and build that infrastructure themselves. 

> Duplicate of #307?

I think this is the alternative to providing additional build targets and arguably a better alternative but requires more work.

ApostolosB:
[...]

PR #607, opened 2022-01-20
Add build target for vkcube-wayland

Tony-LunarG:
Attempts to address #590

### DIFF
--- a/cube/CMakeLists.txt
+++ b/cube/CMakeLists.txt
@@ -69,53 +69,6 @@ if(UNIX AND NOT APPLE) # i.e. Linux
         find_package(WaylandScanner REQUIRED)
         find_package(WaylandProtocols REQUIRED)
         include_directories(${WAYLAND_CLIENT_INCLUDE_DIR})
-    endif()
-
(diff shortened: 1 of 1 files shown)


cube/CMakeLists.txt
```diff
@@ -318,3 +324,36 @@ if(APPLE)
...
+            target_link_libraries(vkcube-wayland Vulkan::Vulkan)
+            target_compile_definitions(vkcube-wayland PUBLIC VK_USE_PLATFORM_WAYLAND_KHR)
+            include(CheckLibraryExists)
+            CHECK_LIBRARY_EXISTS("rt" clock_gettime "" NEED_RT)
+            if (NEED_RT)
+                target_link_libraries(vkcube-wayland rt)
+            endif()
+        endif()
+    endif()
+endif()
```
charles-lunarg, line 359: I wonder if this would be simpler if you created a single library that was the main code but then created 2 executables, one for wayland directly and one that is the existing `vkcube` which will be compiled with the selected windowing system.

This would reduce the code duplication going on here, which I can accept but don't particularly like.

charles-lunarg (approved):
Looks good.

PR #609, opened 2022-01-21
mockicd: Fix FreeCommandBuffers leak

nyorain:
The handles allocated for command buffers were not ever freed.


nyorain:
Hm, I just realized that this does not fix the leak when only `vkDestroyCommandPool` is called (in which case the command buffers are supposed to be freed automatically).

charles-lunarg (approved):
Definitely useful to not leak memory, since it allows mock ICD to be run with memory tools like address sanitizer and not get false positives.

PR #610, opened 2022-01-21
mockicd: Fix queue destruction for multiple devices

nyorain:
vkDestroyDevice would previously clear the queue map which is not correct when there are multiple devices alive.


charles-lunarg (approved):
Yep this looked plain wrong. I appreciate the fix.

PR #613, opened 2022-02-01
 Fix padding for 1.3 promotion of integer dot product, remove malloc from pNext chains.

charles-lunarg:
The 1.3 update forgot to rename
VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_INTEGER_DOT_PRODUCT_FEATURES_KHR to
VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_INTEGER_DOT_PRODUCT_FEATURES as this
is the non-aliased name. Without it, the fix wouldn't be applied.

Also:
Generate a structure that contains all the structures in the pNext chain as
members rather than relying on malloc to allocate memory for them. This makes the code
much less brittle and easier to extend in the future.

### DIFF
--- a/scripts/vulkaninfo_generator.py
+++ b/scripts/vulkaninfo_generator.py
@@ -105,11 +105,12 @@ predefined_types = ['char', 'VkBool32', 'uint32_t', 'uint8_t', 'int32_t',
                     'float', 'uint64_t', 'size_t', 'VkDeviceSize', 'int64_t']
 
 # Types that need pNext Chains built. 'extends' is the xml tag used in the structextends member. 'type' can be device, instance, or both
-EXTENSION_CATEGORIES = OrderedDict((('phys_device_props2', {'extends': 'VkPhysicalDeviceProperties2', 'type': 'both'}),
-                                   ('phys_device_mem_props2', {'extends': 'VkPhysicalDeviceMemoryProperties2', 'type': 'device'}),
-                                   ('phys_device_features2', {'extends': 'VkPhysicalDeviceFeatures2,VkDeviceCreateInfo', 'type': 'device'}),
-                                   ('surface_capabilities2', {'extends': 'VkSurfaceCapabilities2KHR', 'type': 'both'}),
-                                   ('format_properties2', {'extends': 'VkFormatProperties2', 'type': 'device'})
+EXTENSION_CATEGORIES = OrderedDict((
+    ('phys_device_props2', {'extends': 'VkPhysicalDeviceProperties2', 'type': 'both', 'holder_type': 'VkPhysicalDeviceProperties2'}),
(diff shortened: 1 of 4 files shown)


scripts/vulkaninfo_generator.py
```diff
@@ -610,7 +610,7 @@ def PrintChainBuilders(listName, structures, all_structures):
                 # We need to artificially increase it just so the driver doesn't write 'out of bounds' and cause
                 # difficult to debug crashes. This bug comes from the in-development version of the extension having
                 # a larger size than the final version, so older drivers try to writ to members which don't exist.
-                if s.sTypeName == "VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_INTEGER_DOT_PRODUCT_FEATURES_KHR":
+                if s.sTypeName == "VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_INTEGER_DOT_PRODUCT_FEATURES":
```
ziga-lunarg: Should there be a check for both of these?

vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -2326,161 +2326,529 @@ void DumpVkSurfaceProtectedCapabilitiesKHR(Printer &p, std::string name, VkSurfa
...
+        for(size_t i = 0; i < chain_members.size() - 1; i++){
+            chain_members[i]->pNext = chain_members[i + 1];
+        }
+        if (chain_members.size() > 0) start_of_chain = chain_members[0];
     };
-    infos.phys_device_mem_props2 = {
-        {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MEMORY_BUDGET_PROPERTIES_EXT, sizeof(VkPhysicalDeviceMemoryBudgetPropertiesEXT)},
+};
+void setup_phys_device_props2_chain(VkPhysicalDeviceProperties2& start, std::unique_ptr<phys_device_props2_chain>& chain){
+    chain = std::unique_ptr<phys_device_props2_chain>(new phys_device_props2_chain());
```
ziga-lunarg, line 2478: `std::unique_ptr` can be created without using `new`

ziga-lunarg (approved):

scripts/vulkaninfo_generator.py
charles-lunarg: No, because the xml uses only the `SHADER_INTEGER_DOT_PRODUCT_FEATURES` and the KHR version will only appear as an alias, not a distinct structure type.

vulkaninfo/generated/vulkaninfo.hpp
charles-lunarg: Yes, but this repo is still C++11, and `std::make_unique` was added in C++14. Unless you meant a different method.

PR #615, opened 2022-02-03
Cleanup of vkcubepp

### LINKED ISSUE #409: vkcubepp --help has poor formatting
When executing vkcubepp --help:
```
lunarg@lunarg-amd:~$ vkcubepp --help
Usage:
  vkcube	[--use_staging] [--validate]
	[--break] [--c <framecount>] [--suppress_popups]
	[--present_mode <present mode enum>]
	<present_mode_enum>
		VK_PRESENT_MODE_IMMEDIATE_KHR = 0
		VK_PRESENT_MODE_MAILBOX_KHR = 1
		VK_PRESENT_MODE_FIFO_KHR = 2
		VK_PRESENT_MODE_FIFO_RELAXED_KHR = 3lunarg@lunarg-amd:~$ 
```
Note in particular:

* The `Usage` statement refers to `vkcube`, not `vkcubepp`
* The final line doesn't have a newline (which is why the next command prompt appeared at the end)

### LINKED ISSUE #421: cube.cpp misleading use of vulkan.hpp
This is just one example of many:

cube.cpp:
```
    // Get the list of VkFormat's that are supported:
    uint32_t formatCount;
    auto result = gpu.getSurfaceFormatsKHR(surface, &formatCount, static_cast<vk::SurfaceFormatKHR *>(nullptr));
    VERIFY(result == vk::Result::eSuccess);

    std::unique_ptr<vk::SurfaceFormatKHR[]> surfFormats(new vk::SurfaceFormatKHR[formatCount]);
    result = gpu.getSurfaceFormatsKHR(surface, &formatCount, surfFormats.get());
    VERIFY(result == vk::Result::eSuccess);
```

Should be instead:
```
auto surfFormats = gpu.getSurfaceFormatsKHR(surface);
```

As is, the example code is confusing and more likely drives people away from vulkan.hpp instead of showing how useful it really is.
Nobody should do `std::unique_ptr<vk::SurfaceFormatKHR[]> surfFormats(new vk::SurfaceFormatKHR[formatCount]);` when vulkan.hpp can return `std::vector<vk::SurfaceFormatKHR>`.

### LINKED ISSUE #422: cube.cpp bad use of VK_FORMAT_UNDEFINED
In cube.cpp there is code
```
    // If the format list includes just one entry of VK_FORMAT_UNDEFINED,
    // the surface has no preferred format.  Otherwise, at least one
    // supported format will be returned.
    if (formatCount == 1 && surfFormats[0].format == vk::Format::eUndefined) {
        format = vk::Format::eB8G8R8A8Unorm;
...
```
which is checking for condition which should never happen. Vulkan specification does not allow `VK_FORMAT_UNDEFINED` among returned formats.

### LINKED ISSUE #452: vkcube and vkcubpp inconsistent in color selection
image

According to @lunarpapillo this was happening in the samples previously on Mesa drivers. I'm using radv with an AMD driver which is close enough that it's probably the same thing. Apparently the fix in samples was to use the same image format across different drivers instead of just picking the first format. Here is the [samples fix](https://github.com/LunarG/VulkanSamples/pull/302) that could be used as a reference to make the fix in cube and cubepp.

charles-lunarg:
Fixes #421 Fixes #422 Fixes #409 Fixes #452

### DIFF
--- a/cube/cube.cpp
+++ b/cube/cube.cpp
@@ -32,8 +32,9 @@
 #include <cstdlib>
 #include <cstring>
 #include <csignal>
-#include <iostream>
+
 #include <sstream>
+#include <iostream>
 #include <memory>
 
 #define VULKAN_HPP_NO_EXCEPTIONS
@@ -50,14 +51,9 @@
 #endif
 
 #define APP_SHORT_NAME "vkcubepp"
-#ifdef _WIN32
-#define APP_NAME_STR_LEN 80
-#endif
 
 // Allow a maximum of two outstanding presentation operations.
-#define FRAME_LAG 2
-
-#define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
+constexpr uint32_t FRAME_LAG = 2;
 
 #ifdef _WIN32
 #define ERR_EXIT(err_msg, err_class)                                          \
@@ -206,8 +202,7 @@ struct SwapchainImageResources {
 };
 
 struct Demo {
-    Demo();
-    void build_image_ownership_cmd(const SwapchainImageResources &);
+    void build_image_ownership_cmd(const SwapchainImageResources &swapchain_image_resource);
     vk::Bool32 check_layers(const std::vector<const char *> &check_names, const std::vector<vk::LayerProperties> &layers);
     void cleanup();
     void destroy_swapchain_related_resources();
@@ -217,7 +212,7 @@ struct Demo {
     void draw_build_cmd(const SwapchainImageResources &swapchain_image_resource);
     void prepare_init_cmd();
     void flush_init_cmd();
-    void init(int, char **);
+    void init(int argc, char **argv);
     void init_connection();
     void init_vk();
     void init_vk_swapchain();
@@ -229,7 +224,7 @@ struct Demo {
     void prepare_descriptor_pool();
     void prepare_descriptor_set();
     void prepare_framebuffers();
-    vk::ShaderModule prepare_shader_module(const uint32_t *, size_t);
+    vk::ShaderModule prepare_shader_module(const uint32_t *code, size_t size);
     vk::ShaderModule prepare_vs();
     vk::ShaderModule prepare_fs();
     void prepare_pipeline();
@@ -253,17 +248,17 @@ struct Demo {
     void create_window();
 #elif defined(VK_USE_PLATFORM_XLIB_KHR)
     void create_xlib_window();
-    void handle_xlib_event(const XEvent *);
+    void handle_xlib_event(const XEvent *event);
     void run_xlib();
 #elif defined(VK_USE_PLATFORM_XCB_KHR)
-    void handle_xcb_event(const xcb_generic_event_t *);
+    void handle_xcb_event(const xcb_generic_event_t *event);
     void run_xcb();
     void create_xcb_window();
 #elif defined(VK_USE_PLATFORM_WAYLAND_KHR)
     void run();
     void create_window();
 #elif defined(VK_USE_PLATFORM_DIRECTFB_EXT)
-    void handle_directfb_event(const DFBInputEvent *);
+    void handle_directfb_event(const DFBInputEvent *event);
     void run_directfb();
     void create_directfb_window();
 #elif defined(VK_USE_PLATFORM_METAL_EXT)
@@ -273,7 +268,7 @@ struct Demo {
     void run_display();
 #endif
 
-    std::string name;  // Name to put on the window/icon
+    std::string name = "vkcubepp";  // Name to put on the window/icon
 #if defined(VK_USE_PLATFORM_WIN32_KHR)
     HINSTANCE connection = nullptr;  // hInstance - Windows Instance
     HWND window = nullptr;           // hWnd - window handle
@@ -374,9 +369,9 @@ struct Demo {
     vk::RenderPass render_pass;
     vk::Pipeline pipeline;
 
-    mat4x4 projection_matrix;
-    mat4x4 view_matrix;
-    mat4x4 model_matrix;
+    mat4x4 projection_matrix = {};
+    mat4x4 view_matrix = {};
+    mat4x4 model_matrix = {};
 
     float spin_angle = 0.0f;
     float spin_increment = 0.0f;
@@ -470,7 +465,7 @@ static void seat_handle_capabilities(void *data, wl_seat *seat, uint32_t caps) {
         wl_pointer_add_listener(demo->pointer, &pointer_listener, demo);
     } else if (!(caps & WL_SEAT_CAPABILITY_POINTER) && demo->pointer) {
         wl_pointer_destroy(demo->pointer);
-        demo->pointer = NULL;
+        demo->pointer = nullptr;
     }
     // Subscribe to keyboard events
     if (caps & WL_SEAT_CAPABILITY_KEYBOARD) {
@@ -478,7 +473,7 @@ static void seat_handle_capabilities(void *data, wl_seat *seat, uint32_t caps) {
         wl_keyboard_add_listener(demo->keyboard, &keyboard_listener, demo);
     } else if (!(caps & WL_SEAT_CAPABILITY_KEYBOARD)) {
         wl_keyboard_destroy(demo->keyboard);
-        demo->keyboard = NULL;
+        demo->keyboard = nullptr;
     }
 }
 
@@ -512,15 +507,7 @@ static void registry_handle_global_remove(void *data, wl_registry *registry, uin
 static const wl_registry_listener registry_listener = {registry_handle_global, registry_handle_global_remove};
 #endif
 
-Demo::Demo() {
-    name = APP_SHORT_NAME;
-
-    memset(projection_matrix, 0, sizeof(projection_matrix));
-    memset(view_matrix, 0, sizeof(view_matrix));
-    memset(model_matrix, 0, sizeof(model_matrix));
-}
-
-void Demo::build_image_ownership_cmd(SwapchainImageResources const &swapchain_image_resource) {
+void Demo::build_image_ownership_cmd(const SwapchainImageResources &swapchain_image_resource) {
     auto result = swapchain_image_resource.graphics_to_present_cmd.begin(
         vk::CommandBufferBeginInfo().setFlags(vk::CommandBufferUsageFlagBits::eSimultaneousUse));
     VERIFY(result == vk::Result::eSuccess);
@@ -614,7 +601,7 @@ void Demo::cleanup() {
 }
 
 void Demo::create_device() {
-    float priorities = {0.0};
+    float priorities = 0.0;
 
     std::vector<vk::DeviceQueueCreateInfo> queues;
     queues.push_back(vk::DeviceQueueCreateInfo().setQueueFamilyIndex(graphics_queue_family_index).setQueuePriorities(priorities));
@@ -642,7 +629,7 @@ void Demo::draw() {
     device.waitForFences(fences[frame_index], VK_TRUE, UINT64_MAX);
     device.resetFences({fences[frame_index]});
 
-    auto acquire_result = vk::Result::eSuccess;
+    vk::Result acquire_result;
     do {
         acquire_result =
             device.acquireNextImageKHR(swapchain, UINT64_MAX, image_acquired_semaphores[frame_index], vk::Fence(), &current_buffer);
@@ -744,7 +731,7 @@ void Demo::draw_build_cmd(const SwapchainImageResources &swapchain_image_resourc
 
     commandBuffer.bindPipeline(vk::PipelineBindPoint::eGraphics, pipeline);
     commandBuffer.bindDescriptorSets(vk::PipelineBindPoint::eGraphics, pipeline_layout, 0, swapchain_image_resource.descriptor_set,
-                                     nullptr);
+                                     {});
     float viewport_dimension;
     float viewport_x = 0.0f;
     float viewport_y = 0.0f;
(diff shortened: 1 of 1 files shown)


cube/cube.cpp
```diff
@@ -193,27 +193,29 @@ static const float g_uv_buffer_data[] = {
...
     vk::DescriptorSet descriptor_set;
-} SwapchainImageResources;
+};
 
 struct Demo {
     Demo();
-    void build_image_ownership_cmd(uint32_t const &);
-    vk::Bool32 check_layers(uint32_t, const char *const *, uint32_t, vk::LayerProperties *);
+    void build_image_ownership_cmd(const SwapchainImageResources &);
+    vk::Bool32 check_layers(const std::vector<const char *> &check_names, const std::vector<vk::LayerProperties> &layers);
```
lunarpapillo: Some of the function signatures use variable names (this one, `destroy_texture()`, `draw_build_cmd()`, others).  Some do not (`init()`, `prepare_shader_module()`).  They should all match, right?

cube/cube.cpp
```diff
@@ -269,79 +273,77 @@ struct Demo {
     void run_display();
 #endif
 
+    std::string name;  // Name to put on the window/icon
```
lunarpapillo: why not initialize this to `APP_SHORT_NAME` here, since all other initialization is happening here?

cube/cube.cpp
```diff
@@ -511,76 +512,17 @@ static void registry_handle_global_remove(void *data, wl_registry *registry, uin
...
-      validate{false},
-      use_break{false},
-      suppress_popups{false},
-      force_errors{false},
-      current_buffer{0},
-      queue_family_count{0} {
-#if defined(VK_USE_PLATFORM_WIN32_KHR)
-    memset(name, '\0', APP_NAME_STR_LEN);
-#endif
     memset(projection_matrix, 0, sizeof(projection_matrix));
```
lunarpapillo: why are these not also initialized above?

cube/cube.cpp
```diff
@@ -511,76 +512,17 @@ static void registry_handle_global_remove(void *data, wl_registry *registry, uin
...
     memset(view_matrix, 0, sizeof(view_matrix));
     memset(model_matrix, 0, sizeof(model_matrix));
 }
 
-void Demo::build_image_ownership_cmd(uint32_t const &i) {
-    auto const cmd_buf_info = vk::CommandBufferBeginInfo().setFlags(vk::CommandBufferUsageFlagBits::eSimultaneousUse);
-    auto result = swapchain_image_resources[i].graphics_to_present_cmd.begin(&cmd_buf_info);
+void Demo::build_image_ownership_cmd(SwapchainImageResources const &swapchain_image_resource) {
+    auto result = swapchain_image_resource.graphics_to_present_cmd.begin(
```
lunarpapillo: These two commands switched order.  It's not clear why.

cube/cube.cpp
```diff
@@ -703,69 +610,58 @@ void Demo::cleanup() {
...
-
-    auto deviceInfo = vk::DeviceCreateInfo()
-                          .setQueueCreateInfoCount(1)
-                          .setPQueueCreateInfos(queues)
-                          .setEnabledLayerCount(0)
-                          .setPpEnabledLayerNames(nullptr)
-                          .setEnabledExtensionCount(enabled_extension_count)
-                          .setPpEnabledExtensionNames((const char *const *)extension_names)
-                          .setPEnabledFeatures(nullptr);
+    float priorities = {0.0};
```
lunarpapillo: The braces around `0.0` seem unnecessary (and I think they'll break on older Linux compilers).

cube/cube.cpp
```diff
@@ -703,69 +610,58 @@ void Demo::cleanup() {
...
+
+    std::vector<vk::DeviceQueueCreateInfo> queues;
+    queues.push_back(vk::DeviceQueueCreateInfo().setQueueFamilyIndex(graphics_queue_family_index).setQueuePriorities(priorities));
```
lunarpapillo: The original calls `setPQueuePriorities()`, while the new calls `setQueuePriorities()`.  I think this is okay, but would prefer to draw your attention to it to double-check.

cube/cube.cpp
```diff
@@ -703,69 +610,58 @@ void Demo::cleanup() {
...
 }
 
 void Demo::draw() {
     // Ensure no more than FRAME_LAG renderings are outstanding
-    device.waitForFences(1, &fences[frame_index], VK_TRUE, UINT64_MAX);
+    device.waitForFences(fences[frame_index], VK_TRUE, UINT64_MAX);
     device.resetFences({fences[frame_index]});
 
-    vk::Result result;
+    auto acquire_result = vk::Result::eSuccess;
```
lunarpapillo: why initialize?  The value here is immediately overwritten.

cube/cube.cpp
```diff
@@ -774,111 +670,101 @@ void Demo::draw() {
...
 
-    auto const passInfo = vk::RenderPassBeginInfo()
-                              .setRenderPass(render_pass)
-                              .setFramebuffer(swapchain_image_resources[current_buffer].framebuffer)
-                              .setRenderArea(vk::Rect2D(vk::Offset2D(0, 0), vk::Extent2D((uint32_t)width, (uint32_t)height)))
-                              .setClearValueCount(2)
-                              .setPClearValues(clearValues);
-
-    auto result = commandBuffer.begin(&commandInfo);
+    auto result = commandBuffer.begin(vk::CommandBufferBeginInfo().setFlags(vk::CommandBufferUsageFlagBits::eSimultaneousUse));
```
lunarpapillo: got to here and my brain hurts.  I'll pick up more later if I have time.

lunarpapillo (changes requested):
oooo, shiny.
Partial review - I didn't complete the whole thing yet, but wanted to get my observations so far to you in a timely manner.

ziga-lunarg:
Should `malloc` and `free` be replace with `new` and `delete`?

cube/cube.cpp
```diff
@@ -2156,136 +1963,124 @@ void Demo::prepare_render_pass() {
...
 }
 
-void Demo::prepare_texture_buffer(const char *filename, texture_object *tex_obj) {
+void Demo::prepare_texture_buffer(const char *filename, texture_object &tex_obj) {
+    vk::SubresourceLayout tex_layout;
     int32_t tex_width;
     int32_t tex_height;
-
-    if (!loadTexture(filename, NULL, NULL, &tex_width, &tex_height)) {
+    if (!loadTexture(filename, NULL, tex_layout, tex_width, tex_height)) {
```
ziga-lunarg: `nullptr` instead of `NULL`

cube/cube.cpp
```diff
@@ -1470,14 +1325,14 @@ void Demo::create_surface() {
 void Demo::init_vk_swapchain() {
     create_surface();
     // Iterate over each queue to learn whether it supports presenting:
-    std::unique_ptr<vk::Bool32[]> supportsPresent(new vk::Bool32[queue_family_count]);
-    for (uint32_t i = 0; i < queue_family_count; i++) {
+    std::vector<vk::Bool32> supportsPresent(queue_props.size());
+    for (uint32_t i = 0; i < supportsPresent.size(); i++) {
```
ziga-lunarg: Comparing different types, also below

cube/cube.cpp
```diff
@@ -1362,32 +1232,20 @@ void Demo::init_vk() {
...
-            if (!strcmp("VK_KHR_portability_subset", device_extensions[i].extensionName)) {
-                extension_names[enabled_extension_count++] = "VK_KHR_portability_subset";
-            }
-            assert(enabled_extension_count < 64);
+    for (const auto &extension : device_extension_return.value) {
+        if (!strcmp(VK_KHR_SWAPCHAIN_EXTENSION_NAME, extension.extensionName)) {
+            swapchainExtFound = 1;
+            enabled_device_extensions.push_back(VK_KHR_SWAPCHAIN_EXTENSION_NAME);
+        }
+        if (!strcmp("VK_KHR_portability_subset", extension.extensionName)) {
```
ziga-lunarg: Could be `else if`

cube/cube.cpp
```diff
@@ -774,111 +670,101 @@ void Demo::draw() {
...
-
-    vk::Rect2D const scissor(vk::Offset2D(0, 0), vk::Extent2D(width, height));
-    commandBuffer.setScissor(0, 1, &scissor);
+
+    commandBuffer.setViewport(0, vk::Viewport()
+                                     .setX(viewport_x)
+                                     .setY(viewport_y)
+                                     .setWidth(static_cast<float>(viewport_dimension))
+                                     .setHeight(static_cast<float>(viewport_dimension))
+                                     .setMinDepth(static_cast<float>(0.0f))
```
ziga-lunarg: Casting `float` to `float`?

cube/cube.cpp
```diff
@@ -511,76 +512,17 @@ static void registry_handle_global_remove(void *data, wl_registry *registry, uin
...
-#endif
     memset(projection_matrix, 0, sizeof(projection_matrix));
     memset(view_matrix, 0, sizeof(view_matrix));
     memset(model_matrix, 0, sizeof(model_matrix));
 }
 
-void Demo::build_image_ownership_cmd(uint32_t const &i) {
-    auto const cmd_buf_info = vk::CommandBufferBeginInfo().setFlags(vk::CommandBufferUsageFlagBits::eSimultaneousUse);
-    auto result = swapchain_image_resources[i].graphics_to_present_cmd.begin(&cmd_buf_info);
+void Demo::build_image_ownership_cmd(SwapchainImageResources const &swapchain_image_resource) {
```
ziga-lunarg: This is east `const` but function declaration is west `const`

cube/cube.cpp
ziga-lunarg: `0.0f`

cube/cube.cpp
```diff
@@ -703,69 +610,58 @@ void Demo::cleanup() {
...
-    if (tex_objs->buffer) device.destroyBuffer(tex_objs->buffer, nullptr);
+    device.freeMemory(tex_objs.mem);
+    if (tex_objs.image) device.destroyImage(tex_objs.image);
+    if (tex_objs.buffer) device.destroyBuffer(tex_objs.buffer);
 }
 
 void Demo::draw() {
     // Ensure no more than FRAME_LAG renderings are outstanding
-    device.waitForFences(1, &fences[frame_index], VK_TRUE, UINT64_MAX);
+    device.waitForFences(fences[frame_index], VK_TRUE, UINT64_MAX);
```
ziga-lunarg: Could use `std::numerical_limits` instead of macros, but that's all over the code

cube/cube.cpp
charles-lunarg: fixed.

cube/cube.cpp
charles-lunarg: Fixed. I thought I couldn't initialize std::string like that, but I was wrong.

cube/cube.cpp
charles-lunarg: They aren't switched order. I am instead creating the createInfo struct inside the function call instead of outside, because I can. (since it takes it by reference not pointer).

cube/cube.cpp
charles-lunarg: Also fixed. `mat4x4 foo = {};` seemed to do the trick.

cube/cube.cpp
charles-lunarg: Yep, stale code. fixed.

cube/cube.cpp
charles-lunarg: Very intentional. The `setQueuePriorities()` function takes an array proxy, meaning it'll create a 'view' on the float object for us, rather than having to get a pointer to the float directly. more 'modern C++' (and more in line with the vulkan.hpp API)

cube/cube.cpp
charles-lunarg: fixed.

cube/cube.cpp
charles-lunarg: Fixed. (and everywhere else I saw NULL.)

cube/cube.cpp
charles-lunarg: `static_cast<uint32_t>(size())` should prevent the size comparison warnings, but I also modified the function a bit, hope that looks a bit cleaner.

cube/cube.cpp
charles-lunarg: agreed. Added in another place where extensions are being checked.

cube/cube.cpp
charles-lunarg: The old code did `(float)value` here a bunch, was just robotically moving to static_cast. Turns out its entirely unnecessary!

cube/cube.cpp
charles-lunarg: Whoops! My own personal style leaked in there. Its back to west const.

cube/cube.cpp
charles-lunarg: Fixed. (bob also pointed it out, so both of you get a cookie for having eagle eyes).

cube/cube.cpp
charles-lunarg: I did try using numeric limits but got hit with the windows.h max macro headache. It might be possible to use numeric limits in the code, but its not worth the hassle of figuring out why the compiler barfs all over max().

charles-lunarg:
@ziga-lunarg @lunarpapillo Thank you both for reviewing, found probably a dozen real mistakes and inconsistencies.

> Should malloc and free be replace with new and delete?

Turns out I missed a spot! I didn't realize this code was there, I moved it over to using the vulkan.hpp functions that return vectors in a single call instead of the 2-call pattern.

cube/cube.cpp
ziga-lunarg: It's a common issue, you can define NOMINMAX and it should be resolved.

ziga-lunarg (approved):

cube/cube.cpp
charles-lunarg: Indeed. I think i'll leave it as this PR is big enough as it is.

PR #618, opened 2022-02-08
vulkaninfo: Enable portability subset extension if present

charles-lunarg:
If this extension is present, it must be enabled when creating a VkDevice.


mikes-lunarg (approved):
LGTM

PR #622, opened 2022-02-09
vulkaninfo: Enable portability extension for device groups

mikes-lunarg:
Expand previous change to include device group creation. Also added some
VK_ENABLE_BETA_EXTENSIONS guards around the portability identifiers.


charles-lunarg (approved):
Good changes. Forgot that portability needs to be guarded by BETA_EXTENSIONS for now.

PR #623, opened 2022-02-09
Make device selection output constistent

jeremy-lunarg:
Make vkcube print a string for device type, which is consistent with
vkcubepp.

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -292,6 +292,19 @@ void dumpVec4(const char *note, vec4 vector) {
     fflush(stdout);
 }
 
+char const* to_string(VkPhysicalDeviceType const type)
+{
+    switch(type)
+    {
+        case VK_PHYSICAL_DEVICE_TYPE_OTHER: return "Other";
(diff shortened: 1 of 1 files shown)


cube/cube.c
```diff
@@ -292,6 +292,19 @@ void dumpVec4(const char *note, vec4 vector) {
...
+char const* to_string(VkPhysicalDeviceType const type)
+{
+    switch(type)
+    {
+        case VK_PHYSICAL_DEVICE_TYPE_OTHER: return "Other";
+        case VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU: return "IntegratedGpu";
+        case VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU: return "DiscreteGpu";
+        case VK_PHYSICAL_DEVICE_TYPE_VIRTUAL_GPU: return "VirtualGpu";
+        case VK_PHYSICAL_DEVICE_TYPE_CPU: return "Cpu";
+        default: return "Invalid";
```
charles-lunarg, line 304: Probably better to say 'Unknown' than invalid, since it could be that there is a new type added and this code wasn't updated.

charles-lunarg (approved):

PR #626, opened 2022-02-23
vulkaninfo: Replace devsim with profile json output

christophe-lunarg:
Initial commit to make the output occur, will need additional refinement.

### DIFF
--- a/scripts/vulkaninfo_generator.py
+++ b/scripts/vulkaninfo_generator.py
@@ -59,20 +59,15 @@ license_header = '''
 '''
 
 custom_formatters = r'''
-void DumpVkConformanceVersion(Printer &p, std::string name, VkConformanceVersion &c) {
-    p.PrintKeyString("conformanceVersion", std::to_string(c.major)+ "." + std::to_string(c.minor) + "." + std::to_string(c.subminor) + "."
-             + std::to_string(c.patch));
-}
-
 template <typename T>
-std::string to_hex_str(T i) {
+std::string to_hex_str(const T i) {
     std::stringstream stream;
     stream << "0x" << std::setfill('0') << std::setw(sizeof(T)) << std::hex << i;
     return stream.str();
 }
 
 template <typename T>
-std::string to_hex_str(Printer &p, T i) {
+std::string to_hex_str(Printer &p, const T i) {
     if (p.Type() == OutputType::json)
         return std::to_string(i);
     else if (p.Type() == OutputType::vkconfig_output)
@@ -89,7 +84,7 @@ structures_to_gen = ['VkExtent3D', 'VkExtent2D', 'VkPhysicalDeviceLimits', 'VkPh
                      'VkSurfaceCapabilitiesKHR', 'VkSurfaceFormatKHR', 'VkLayerProperties', 'VkPhysicalDeviceToolProperties']
 enums_to_gen = ['VkResult', 'VkFormat', 'VkPresentModeKHR',
                 'VkPhysicalDeviceType', 'VkImageTiling']
-flags_to_gen = ['VkSurfaceTransformFlagsKHR', 'VkCompositeAlphaFlagsKHR', 'VkSurfaceCounterFlagsEXT',
+flags_to_gen = ['VkSurfaceTransformFlagsKHR', 'VkCompositeAlphaFlagsKHR', 'VkSurfaceCounterFlagsEXT', 'VkQueueFlags',
                 'VkDeviceGroupPresentModeFlagsKHR', 'VkFormatFeatureFlags', 'VkFormatFeatureFlags2', 'VkMemoryPropertyFlags', 'VkMemoryHeapFlags']
 flags_strings_to_gen = ['VkQueueFlags']
 
@@ -98,11 +93,16 @@ struct_short_versions_to_gen = ['VkExtent3D']
 struct_comparisons_to_gen = ['VkSurfaceFormatKHR', 'VkSurfaceFormat2KHR', 'VkSurfaceCapabilitiesKHR',
                              'VkSurfaceCapabilities2KHR', 'VkSurfaceCapabilities2EXT']
 # don't generate these structures
-struct_blacklist = ['VkConformanceVersion']
+struct_blacklist = ['VkVideoProfilesKHR', 'VkVideoProfileKHR', 'VkDrmFormatModifierPropertiesListEXT', 'VkDrmFormatModifierPropertiesEXT', 'VkDrmFormatModifierPropertiesList2EXT']
+
+# generate these structures such that they only print when not in json mode (as json wants them separate)
+portability_structs = ['VkPhysicalDevicePortabilitySubsetFeaturesKHR', 'VkPhysicalDevicePortabilitySubsetPropertiesKHR']
 
 # iostream or custom outputter handles these types
 predefined_types = ['char', 'VkBool32', 'uint32_t', 'uint8_t', 'int32_t',
                     'float', 'uint64_t', 'size_t', 'VkDeviceSize', 'int64_t']
+
+names_to_ignore = ['sType', 'pNext', 'stdProfileIdc']
 
 EXTENSION_TYPE_INSTANCE = 'instance'
 EXTENSION_TYPE_DEVICE = 'device'
@@ -364,24 +364,24 @@ class VulkanInfoGenerator(OutputGenerator):
                         return
 
         for key, value in EXTENSION_CATEGORIES.items():
-            if typeinfo.elem.get('structextends') == value.get('extends'):
+            if str(typeinfo.elem.get('structextends')).find(value.get('extends')) != -1:
                 self.extension_sets[key].add(name)
 
 
 def GatherTypesToGen(structure_list, structures):
     types = set()
+    for s in structures:
+        types.add(s)
     added_stuff = True  # repeat until no new types are added
     while added_stuff == True:
         added_stuff = False
-        for s in (x for x in structure_list if x.name in structures):
-            size = len(types)
-            types.add(s.name)
-            if len(types) != size:
-                added_stuff = True
-            for m in s.members:
-                if m.typeID not in predefined_types and m.name not in ['sType', 'pNext']:
-                    types.add(m.typeID)
-    types = sorted(types)
+        for s in structure_list:
+            if s.name in types:
+                for m in s.members:
+                    if m.typeID not in predefined_types and m.name not in names_to_ignore:
+                        if m.typeID not in types:
+                            types.add(m.typeID)
+                            added_stuff = True
     return types
 
 
@@ -425,10 +425,10 @@ def PrintEnum(enum, gen):
     out = ''
     out += AddGuardHeader(GetExtension(enum.name, gen))
     out += f"""void Dump{enum.name}(Printer &p, std::string name, {enum.name} value) {{
-    if (p.Type() == OutputType::json) {{
-        p.PrintKeyValue(name, value);
-    }} else {{
-        p.PrintKeyString(name, {enum.name}String(value));\n    }}
+    if (p.Type() == OutputType::json)
+        p.PrintKeyString(name, std::string("VK_") + {enum.name}String(value));
+    else
+        p.PrintKeyString(name, {enum.name}String(value));
 }}
 """
     out += AddGuardFooter(GetExtension(enum.name, gen))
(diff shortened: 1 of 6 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -951,13 +964,11 @@ PrinterCreateDetails get_printer_create_details(ParsedResults &parse_data, AppIn
...
+        case (OutputCategory::profile_json):
             create.output_type = OutputType::json;
-            create.start_string = std::string("{\n\t\"$schema\": \"https://schema.khronos.org/vulkan/devsim_1_0_0.json#\",\n") +
-                                  "\t\"comments\": {\n\t\t\"desc\": \"JSON configuration file describing GPU " +
-                                  std::to_string(parse_data.selected_gpu) + " (" + selected_gpu.props.deviceName +
-                                  "). Generated using the " + executable_name + " program.\",\n\t\t\"vulkanApiVersion\": \"" +
-                                  VkVersionString(inst.vk_version) + "\"\n" + "\t}";
+            create.start_string =
+                "{\n\t\"$schema\": "
+                "\"https://raw.githubusercontent.com/KhronosGroup/Vulkan-Profiles/master/schema/profile_schema.json\"";
```
christophe-lunarg, line 971: The profile file is not generated for any version of the profile schema but for the profile schema API version that matches a Vulkan Header revision ideally.

This allows avoiding to require Vulkan capabilities that didn't exist for this Vulkan Header revision.

christophe-lunarg (commented):
The capabilities section is only a container for capabilities. 

It's only the profile section that says what a profile is made of. Vulkan Profiles can be defined within a file so that a file may express multiple variants of a profile.

However the Vulkaninfo output a specific kind of Vulkan Profiles: device (and driver specific) profile so there will be a single profile per file, this said with the portability extensions this may be different.

Note as well that a profile also specific how a Vulkan capability should be used. So if the driver reports Vulkan 1.2.198, the generated profile file should not rely on Vulkan 1.3 core structures.

You can use https://www.jsonschemavalidator.net/ and schemas in https://schema.khronos.org/vulkan/ to check the files are generated correctly.

christophe-lunarg:
Currently, we only have schema versions for each SDK revision we shipped since 1.1.106 but I think we should generate a schema for each Vulkan Header version we can.

It's probably not more than a day of work.

However, it raises the question of uploading Vulkan schemas for each Vulkan header revision. Do we follow the rythmn of Vulkan Header revisions, do we do it during SDK release?

vulkaninfo/vulkaninfo.cpp
charles-lunarg: So this should be changed such that the schema used has the same API version as the device?
Also is there going to be an official location for these schema files? IE currently i'm referencing this repo but that doesn't seem very official.

lunarpapillo:
FYI, the last CI failure seems to be due to a CMake 3.19 bug.  I'm rerunning.

charles-lunarg:
@christophe-lunarg I have gotten this PR into a good shape. It passes the schema and the vulkan-profiles repo can build with the output json file. The Portability Subset is handled like the `VP_LUNARG_desktop_portability_2021.json`, where it creates a separate capability group and profile which includes the baseline and portability stuff.

scripts/vulkaninfo_generator.py
```diff
@@ -98,12 +93,17 @@
 struct_comparisons_to_gen = ['VkSurfaceFormatKHR', 'VkSurfaceFormat2KHR', 'VkSurfaceCapabilitiesKHR',
                              'VkSurfaceCapabilities2KHR', 'VkSurfaceCapabilities2EXT']
 # don't generate these structures
-struct_blacklist = ['VkConformanceVersion']
+struct_blacklist = ['VkVideoProfilesKHR']
```
aqnuep, line 96: The generator script also disallow-lists the various DRM format modifier list structures for the same reason as `VkVideoProfilesKHR`, i.e. because they contain pointers to arrays which shouldn't be used in output structures. In order to stay consistent with the Vulkan-Profiles toolchain, you probably should do the same here so that you don't accidentally generate profile JSONs containing those disallow-listed structures, as they won't comply with the schema or work with the library generator.

aqnuep (approved):
The changes look fine in general, but it's impossible to tell if anything is missing or wrong just looking at the code, so in the end what matters is that this is covered with sufficient level of testing.

charles-lunarg:
There are a couple of things which aren't solidified yet.

*  The contributors section. I left this blank for now, what should go there? LunarG? Khronos? Anyone whose ever edited vulkaninfo?
* I found on hardware that supports the video extension the VkFormats blocks would contain empty sections for `VkVideoProfileKHR`. This may be from what @aqnuep described, that some structs shouldn't be printed.
* A portability device will output the portability data (features, extension, properties) in a separate capabilities block and creates a separate profile for it. Is this an appropriate way to handle portability devices? (let me know if an example json should be uploaded).

vulkaninfo/vulkaninfo.cpp
```diff
@@ -40,17 +40,20 @@ void DumpExtensions(Printer &p, std::string layer_name, std::vector<VkExtensionP
     for (const auto &ext : extensions) {
         max_length = std::max(max_length, std::strlen(ext.extensionName));
     }
-
-    ObjectWrapper obj(p, layer_name + " Extensions", extensions.size());
+#if defined(VK_ENABLE_BETA_EXTENSIONS)
+    std::string portability_ext_name = VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME;
```
ziga-lunarg, line 44: Could be const

vulkaninfo/vulkaninfo.cpp
```diff
@@ -40,17 +40,20 @@ void DumpExtensions(Printer &p, std::string layer_name, std::vector<VkExtensionP
...
+#endif  // defined(VK_ENABLE_BETA_EXTENSIONS)
+    ObjectWrapper obj(p, section_name, extensions.size());
     if (do_indent) p.IndentDecrease();
     for (auto &ext : extensions) {
```
ziga-lunarg, line 48: const

vulkaninfo/vulkaninfo.cpp
```diff
@@ -513,41 +464,33 @@ void GpuDumpFeatures(Printer &p, AppGpu &gpu) {
 }
 
 void GpuDumpFormatProperty(Printer &p, VkFormat fmt, VkFormatProperties prop) {
+    std::string name = "";
```
ziga-lunarg, line 467: Initializing with `{}` is better

vulkaninfo/vulkaninfo.cpp
```diff
@@ -664,31 +589,177 @@ void DumpGpu(Printer &p, AppGpu &gpu, bool show_tooling_info, bool show_formats)
...
+                        gpu.inst.dll.fp_vkGetPhysicalDeviceFormatProperties(gpu.phys_device, fmt, &props);
+
+                        // don't print format properties that are unsupported
+                        if ((props.linearTilingFeatures || props.optimalTilingFeatures || props.bufferFeatures) == 0) continue;
+
+                        ObjectWrapper format_obj(p, std::string("VK_") + VkFormatString(fmt));
+                        {
+                            GpuDumpFormatProperty(p, fmt, props);
+
+                            VkFormatProperties2 format_props2;
```
ziga-lunarg, line 642: Not sure if I'm missing something, but I don't see where `sType` is set. `pNext` is set in `setup_format_properties2_chain`, but I would also like to see it set to `nullptr` here for clarity.

vulkaninfo/vulkaninfo.cpp
```diff
@@ -664,31 +589,177 @@ void DumpGpu(Printer &p, AppGpu &gpu, bool show_tooling_info, bool show_formats)
...
-            p.PrintExtension(ext.extensionName, ext.specVersion);
+#if defined(VK_ENABLE_BETA_EXTENSIONS)
+    // Print portability subset extension, features, and properties if available
+    if (gpu.CheckPhysicalDeviceExtensionIncluded(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME) &&
+        (gpu.inst.CheckExtensionEnabled(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME) ||
+         gpu.inst.instance_version >= VK_API_VERSION_1_1)) {
+        ObjectWrapper macos_obj(p, "macos-specific");
+        {
+            ObjectWrapper ext_obj(p, "extensions");
+            std::string portability_ext_name = VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME;
```
ziga-lunarg, line 677: const

vulkaninfo/vulkaninfo.cpp
```diff
@@ -664,31 +589,177 @@ void DumpGpu(Printer &p, AppGpu &gpu, bool show_tooling_info, bool show_formats)
...
-    GpuDevDumpJson(p, gpu);
+// Prints profiles section of profiles schema
+void DumpGpuProfileInfo(Printer &p, AppGpu &gpu) {
+    ObjectWrapper profiles(p, "profiles");
+
+    std::string device_label = std::string(gpu.props.deviceName) + " driver " + VkVersionString(gpu.props.driverVersion);
+    std::string device_name =
+        std::string("VP_VULKANINFO_") + std::string(gpu.props.deviceName) + "_" + VkVersionString(gpu.props.driverVersion);
+    ;
+    for (auto &c : device_name) {
```
ziga-lunarg, line 745: I always like to make references const

aqnuep:
> I found on hardware that supports the video extension the VkFormats blocks would contain empty sections for VkVideoProfileKHR. This may be from what @aqnuep described, that some structs shouldn't be printed.

Yes, that may be related. We need to clarify intent on this one.

> A portability device will output the portability data (features, extension, properties) in a separate capabilities block and creates a separate profile for it. Is this an appropriate way to handle portability devices? (let me know if an example json should be uploaded).

I'd say that if we want to generate separate portability fallback profiles in the same profile JSON for each device then we might have to follow a similar pattern to the desktop portability profiles, i.e. have a separate capability set for the portability specific part and for stuff that cannot be in portability. This latter is important as e.g. some features like robust buffer access are not required in portability.

However, I'm not sure whether it's needed to generate a portability-compatible profile as well. That's open for debate I think.

vulkaninfo/vulkaninfo.cpp
christophe-lunarg: Yes same API version than the device.

The official location: https://schema.khronos.org/vulkan/

charles-lunarg: I can't use the same schema as the device version because it may not be on the website.
For example, MoltenVK uses 1.1.304.

vulkaninfo/vulkaninfo.cpp
charles-lunarg: `setup_format_properties2_chain` will set the pnext pointer. I changed the declaration to `VkFormatProperties2 format_props2{};` so it should be null'd out there.

scripts/vulkaninfo_generator.py
charles-lunarg: The struct blacklist now contains the following.
```python
struct_blacklist = ['VkVideoProfilesKHR', 'VkVideoProfileKHR', 'VkDrmFormatModifierPropertiesListEXT', 'VkDrmFormatModifierPropertiesEXT', 'VkDrmFormatModifierPropertiesList2EXT']
```

charles-lunarg:
> I'd say that if we want to generate separate portability fallback profiles in the same profile JSON for each device then we might have to follow a similar pattern to the desktop portability profiles, i.e. have a separate capability set for the portability specific part and for stuff that cannot be in portability.

This is what I have done currently. If running on a portability device, there will be a `macos-specific` capabilities object with the portability features, properties, and extension, while the `device` capabilities will omit those features/properties/extensions. Whether is necessary I will let others decide.

charles-lunarg:
This PR is waiting on a final decision on how to handle the schema versioning. Currently the output uses a `profiles-1.x.y.json` but that has difficulties with systems that don't support the latest minor version with the latest patch version (such as MoltenVK).

christophe-lunarg:
Looks great!

charles-lunarg:
I'm going to merge this shortly (next day or two) so this is a reminder to any last minute reviews. I think the PR is in a good state. I don't recall there being any tasks left to do.

christophe-lunarg:
Yes please!

charles-lunarg:
Currently the output file is named `vulkaninfo.json`. Is this the default name we want? vulkan.gpuinfo.org prints the driver name and date in the filename.

christophe-lunarg:
"vulkan.gpuinfo.org prints the driver name and date in the filename"

True, this is a good idea.

PR #639, opened 2022-04-06
build: Factor out typemap_helper header in rules

lokokung:
For reusability in Dawn without depending on entire icd lib, it is helpful to factor out the header into a separate source set so that we don't need to depend on the shared library.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -46,17 +46,25 @@ config("vulkan_internal_config") {
   }
 }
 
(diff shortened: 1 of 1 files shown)


lokokung:
@charles-lunarg Just tagging an active reviewer since I can't seem to add any reviewers for this atm. Thanks in advance!

charles-lunarg (approved):

PR #641, opened 2022-04-11
cube,vulkaninfo: Add portability_enumeration if present

charles-lunarg:
Make vulkaninfo and vkcube/vkcubepp enable the portability enumeration bit if it is
available.


Tony-LunarG (approved):
LGTM - thanks for the comment

PR #642, opened 2022-04-12
build: Guard chromium-specific configs in GN

brianosman:
Skia is going to start using these GN files, but does not have chromium's //build. To get them to build, guard those references using `build_with_chromium`. A similar mechanism (with the same argument) is used throughout other third party dependencies.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import("//build_overrides/build.gni")
 import("//build_overrides/vulkan_tools.gni")
 
 # Vulkan-tools isn't ported to Fuchsia yet.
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
Looks fine to me.

charles-lunarg:
Seems that the GN CI build here failed - it mentions an unannotated fallthrough in the MockICD code. This should probably be fixed rather than ignoring the warning/error.

brianosman:
Yes, I wasn't expecting these failures, but I think that my change happens to alter which compile flags are used on the CI (chromium defaults to using more restrictive flags for "chromium_code" than "no_chromium_code"). The standalone build here happens to not set "build_with_chromium", so we're getting the stricter rules. It looks like this library only requires C++14, so [[fallthrough]] isn't available?

charles-lunarg:
`[[clang:fallthrough]]` appears to not be liked due to it being unrecognized. Since the use of fallthroughs here isn't really necessary, its easier to remove them outright, as I have done in #643. I'd greatly appreciate it if you could verify that the change works for you.

brianosman:
Yes, I patched that in, and it works fine. Much cleaner solution, thank you!

charles-lunarg:
The CI is running without the commit with the fix. Should just need to rebase and push.

PR #652, opened 2022-05-18
Include crtdbg.h for _CRT_ASSERT

xantares:


charles-lunarg (commented):
While this change looks reasonable, could I get a quick summary of why the change is needed? Possibly put as a comment in the code, for future reference.

charles-lunarg:
Ah well this behavior is much clearer. But from a code cleanliness perspective, its better for remove the work-in-progress commit, just leaving the actually intended changes.

charles-lunarg (approved):

charles-lunarg:
Seems that I missed a `)` while looking at the code. Just fix that and as long as it passes CI, I will merge it.

xantares:
oh yes

PR #653, opened 2022-05-19
cube: add null check on wl_*_destroy functions

tleydxdy:
Encountered the crash on a compositor without keyboard, this seems to
fix it.


charles-lunarg (commented):
Changes look good, but the CLA will need to be signed before we can merge this.

PR #656, opened 2022-05-26
vkcubepp: Use correct extension for portability enumeration

charles-lunarg:
Used the wrong extension name by accident.


mikes-lunarg (approved):
LGTM

PR #661, opened 2022-06-14
win32: use _set_abort_behavior from ucrt

xantares:
turns out it is also available on mingw


charles-lunarg (approved):
I'll merge assuming they pass internal CI.

PR #669, opened 2022-07-18
Set usage on transient image based on format

jfthibert:
Avoids setting the color attachment bit for depth stencil formats to prevent crash


jfthibert:
We were seeing the issue on slightly older versions of amdvlk ICD where an update of vulkaninfo caused the crashes, we haven't seen that behavior on any other application so far. Are there any cases where it would make sense to have depth format also be used as color attachment? I agree the logic is slightly hard to follow, it might make sense to extract additional usage bits for transient outside of the tiling check?

charles-lunarg (approved):
On re-reading the code, I believe this PR is good, since the `VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT` should only be applied to color image formats (which is done so later on line 1636).

charles-lunarg:
Since you are seeing real crashes, and since color & depth_stencil usages are usually mutually exclusive, I'm going to merge it.

PR #670, opened 2022-07-21
vulkaninfo: include wayland-client because its functions are used

jengelh:
"Include what you use": `vulkaninfo.h` calls `wl_registry_bind`, thus
ought to include `wayland-client.h`.

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -79,6 +79,10 @@
 #include "metal_view.h"
 #endif
 
+#if defined(VK_USE_PLATFORM_WAYLAND_KHR)
+#include <wayland-client.h>
+#endif
+
 #include <vulkan/vulkan.h>
 
 static std::string VkResultString(VkResult err);


charles-lunarg:
This isn't a needed change, because wayland-client.h is automatically included by the vulkan.h header when `VK_USE_PLATFORM_WAYLAND_KHR` is defined. Unless there is an benefit of this change that I am unaware of, I'm inclined to decline this PR.

jengelh:
Given an input source of just two lines
```
#define VK_USE_PLATFORM_WAYLAND_KHR
#include <vulkan/vulkan.h>
```
then building this with
```
gcc -c x.c `pkg-config vulkan --cflags`
```
can fail, because vulkan.pc is missing a `Requires: wayland-client` and so, vital `-I` flags may be absent. I can sense such a Require is unpopular, since it would add wayland to everyone.

One path I explored is to drop `#include <wayland-client.h>` from vulkan.h, since vulkan.h only uses forward-declarable struct names from wl-c currently.

And that's why [IWYU](https://github.com/include-what-you-use/include-what-you-use) is a good idea:
"a state where every file includes the headers it needs to declare the symbols that it uses. When every file includes what it uses, then it is possible to edit any file and remove unused headers, without fear of accidentally breaking the upwards dependencies"

cubanismo:
Note we are also trying to get vulkan.h to stop including wayland-client.h, because it doesn't actually need it, it dumps a ton of warnings in our driver builds, and it can lead to mistakes like this. IWYU is indeed a good guiding principle.

charles-lunarg:
Ah that was some context I wasn't aware of. In that case, I definitely can take the PR. 
Thanks for the clarification!

PR #671, opened 2022-07-21
build: Update to header 1.3.222

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):
Vendor only extensions added this time around, so not many changes to worry about.

mikes-lunarg:
Now with actual changes!

PR #679, opened 2022-08-31
Move BUILD.gn values into the scope that uses them

toji:
There are several values that in the build file that are only used when
the platform is not Android, but which are declared outside the
`if(!is_android)` block. This causes unused variable errors when
attempting to build chromium on Android with a dependency on this repo.

By moving the definitions into the conditional block the error is
resolved.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -20,19 +20,6 @@ import("//build_overrides/vulkan_tools.gni")
 # TODO(spang): Port mock ICD to Fuchsia.
 assert(!is_fuchsia)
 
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):

toji:
Apologies, I'm not familiar with the PR workflow on this repo. Are there any further steps I need to take prior to this being merged?

charles-lunarg:
Nothing, I just was waiting for CI to finish. And then promptly went off to do something else, forgetting to check in on this PR.

PR #684, opened 2022-09-22
build: Update to header 1.3.229

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):
Looks like a 'no functionality change-change', so good to go.

PR #691, opened 2022-10-19
cubepp: Fix presentKHR assert

juan-lunarg:
Currently cubepp fails on Debug builds and triggers an assert.

This avoids the assert by calling the appropriate presentKHR.

### DIFF
--- a/cube/cube.cpp
+++ b/cube/cube.cpp
@@ -706,14 +706,15 @@ void Demo::draw() {
         VERIFY(change_owner_result == vk::Result::eSuccess);
     }
 
+    const auto presentInfo = vk::PresentInfoKHR()
+                                 .setWaitSemaphores(separate_present_queue ? image_ownership_semaphores[frame_index]
+                                                                           : draw_complete_semaphores[frame_index])
+                                 .setSwapchains(swapchain)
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
Looks great.
Just need to get CI to run

cube/cube.cpp
```diff
@@ -706,14 +706,15 @@ void Demo::draw() {
...
+
     // If we are using separate queues we have to wait for image ownership,
     // otherwise wait for draw complete
-    auto present_result =
-        present_queue.presentKHR(vk::PresentInfoKHR()
-                                     .setWaitSemaphores(separate_present_queue ? image_ownership_semaphores[frame_index]
-                                                                               : draw_complete_semaphores[frame_index])
-                                     .setSwapchains(swapchain)
-                                     .setImageIndices(current_buffer));
+    auto present_result = present_queue.presentKHR(&presentInfo);
```
juan-lunarg, line 717: By passing a pointer instead of a const& we are invoking the `presentKHR` function that **doesn't** assert when the return code isn't successful.

Which is what our code expects.

juan-lunarg:
I can't merge this change @charles-lunarg I don't have write access for this repo

PR #694, opened 2022-10-26
Cmake cleanup

juan-lunarg:
I spread this into several atomic commits. Worth noting for review purposes.


cube/CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ else()
 endif()
 
 if(UNIX AND NOT APPLE) # i.e. Linux
-    include(FindPkgConfig)
+    find_package(PkgConfig)
```
charles-lunarg, line 52: Why have find_package here and not in a more common place? (it looks like it is present here but not in the vulkaninfo CMakeLists.

charles-lunarg (approved):
Just a comment, otherwise the changes look good.

cube/CMakeLists.txt
juan-lunarg: Because I am error.

juan-lunarg: I thought cube/CMakeLists.txt was the top level CMakeLists.txt not sure why

juan-lunarg: Fixed now.

PR #695, opened 2022-10-26
cmake: Sync FindVulkanHeaders.cmake with VVL

mikes-lunarg:
This picks up a few fixes, notably https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1393 which helps with cross-compiling.


juan-lunarg (approved):
LGTM

charles-lunarg (approved):

PR #699, opened 2022-10-28
Vulkaninfo: Escape json strings

### LINKED ISSUE #696: vulkaninfo --json needs to escape special characters inside strings
Similar to issue reported here: https://github.com/LunarG/gfxreconstruct/issues/847

`vulkaninfo --json` generates an invalid JSON file if special characters such as `\n` or `"` are present inside a string:

```
...
"VkPhysicalDeviceDriverPropertiesKHR": {
	"driverID": "VK_DRIVER_ID_QUALCOMM_PROPRIETARY",
	"driverName": "Qualcomm Technologies Inc. Adreno Vulkan Driver",
	"driverInfo": "Driver Build: 85da404, I46ff5fc46f, 1606794520
Date: 11/30/20
Compiler Version: EV031.31.04.01
Driver Branch: promo490_3_Google
",
	"conformanceVersion": {
		"major": 1,
		"minor": 2,
		"subminor": 0,
		"patch": 1
	}
},
...
```

charles-lunarg:
Fix #696 

Also:
* Parse & print `VkPhysicalDeviceProperties:driverVersion' according to driver specific formats. Falls back on the Vulkan Major.Minor.Patch if on an unknown platform.
* Switch the `VkPhysicalDeviceProperties:apiVersion` printing to put the major.minor.patch outside of the parenthesis. `apiVersion = 1.3.245 (232343235)` is how it is now printed, instead of ` apiVersion = 232343235 (1.3.245)
* Cleanup the logic around printing UUID arrays. Took some effort but uses the operator<< rather than requiring explicit conversion to strings.

### DIFF
--- a/vulkaninfo/outputprinter.h
+++ b/vulkaninfo/outputprinter.h
@@ -51,20 +51,9 @@ std::string to_string(const std::array<uint8_t, 8> &uid) {
     return ss.str();
 }
 
-std::string VkVersionString(uint32_t version) {
-    uint32_t major = version >> 22;
-    uint32_t minor = (version >> 12) & 0x3ff;
-    uint32_t patch = version & 0xfff;
-    return std::to_string(major) + "." + std::to_string(minor) + "." + std::to_string(patch);
-}
-
-std::string VkVersionString(VulkanVersion v) {
-    return std::to_string(v.major) + "." + std::to_string(v.minor) + "." + std::to_string(v.patch);
-}
-
-std::string VkConformanceVersionString(const VkConformanceVersion &c) {
-    return std::to_string(c.major) + "." + std::to_string(c.minor) + "." + std::to_string(c.subminor) + "." +
-           std::to_string(c.patch);
+std::ostream &operator<<(std::ostream &out, const VkConformanceVersion &v) {
+    return out << static_cast<unsigned>(v.major) << "." << static_cast<unsigned>(v.minor) << "."
+               << static_cast<unsigned>(v.subminor) << "." << static_cast<unsigned>(v.patch);
 }
 
 enum class OutputType { text, html, json, vkconfig_output };
@@ -88,7 +77,7 @@ class Printer {
                 out << "==========\n";
                 out << "VULKANINFO\n";
                 out << "==========\n\n";
-                out << "Vulkan Instance Version: " << VkVersionString(vulkan_version) << "\n\n\n";
+                out << "Vulkan Instance Version: " << VulkanVersion(vulkan_version) << "\n\n\n";
 
                 break;
             case (OutputType::html):
@@ -180,7 +169,7 @@ class Printer {
                 out << "\t\t\t<h1>vulkaninfo</h1>\n";
                 out << "\t\t</div>\n";
                 out << "\t\t<div id='wrapper'>\n";
-                out << "\t\t\t<details><summary>Vulkan Instance Version: <span class='val'>" << VkVersionString(vulkan_version)
+                out << "\t\t\t<details><summary>Vulkan Instance Version: <span class='val'>" << VulkanVersion(vulkan_version)
                     << "</span></summary></details>\n\t\t\t<br />\n";
                 node.indents = 3;
                 break;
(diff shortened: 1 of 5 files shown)


charles-lunarg:
Added bob as a reviewer since he would like to know about any changes to the format vulkaninfo uses.

vulkaninfo/outputprinter.h
```diff
@@ -51,20 +51,28 @@ std::string to_string(const std::array<uint8_t, 8> &uid) {
...
 }
 
-std::string VkVersionString(VulkanVersion v) {
-    return std::to_string(v.major) + "." + std::to_string(v.minor) + "." + std::to_string(v.patch);
-}
-
-std::string VkConformanceVersionString(const VkConformanceVersion &c) {
-    return std::to_string(c.major) + "." + std::to_string(c.minor) + "." + std::to_string(c.subminor) + "." +
-           std::to_string(c.patch);
+// Vendor specific driverVersion mapping scheme
```
mikes-lunarg, line 59: I'm iffy on this. It is arguably more correct key off `VkPhysicalDeviceDriverProperties::driverID` than `VkPhysicalDeviceProperties::vendorID`, but that struct already has a string field with the human-readable driver version. I'm assuming mesa NVK driver will report the NVIDIA vendor ID and get a mangled version as a result.

charles-lunarg: Thats an excellent point. I think using VkPhysicalDeviceDriverProperties::driverID is superior just because it allows more specificity. I would likely need a fallback path for 1.0 drivers though, which means this code isn't completely useless.

As far as the NVK driver goes, I think using driverID will hopefully circumvent this issue. I assume NVK will use the Mesa versioning scheme, not the NVidia one, so using the Vulkan API version would be correct.

FYI the driverVersion formatting was taken from https://github.com/SaschaWillems/vulkan.gpuinfo.org/blob/cc83072c855ea00ad4c5f85148e822c094668fcf/includes/functions.php#L364

mikes-lunarg: That's an interesting snippet from gpuinfo.org and it probably predates VK_KHR_driver_properties. I would still be inclined to scrap the uint32_t -> semantic version string conversion because 1) There isn't enough information to do it accurately without VkPhysicalDeviceDriverProperties(KHR) and 2) Given VkPhysicalDeviceDriverProperties, the driver has already done a string conversion for you. I guess that means I'm ok with old drivers have semi-opaque versions, that was probably what motivated the introduction of VK_KHR_driver_properties in the first place.

lunarpapillo: I think this refers to changing what was historically an integer field to a string field... I'm against this in principle, as it's a unnecessary compatibility break.  Adding a new field with a formatted/mangled/whatever value is more acceptable than breaking downstream users who are trying to use `vulkaninfo -j` for production (like, our CI, which will certainly break).

charles-lunarg: I just double checked the schema, and changing the apiVersion & driverVersion in the profiles output fails the schema, so I'll definitely be changing those fields back to using the uint32_t's instead of the human readable strings.

charles-lunarg: @mikes-lunarg Something to note about `VkPhysicalDeviceDriverProperties::driverInfo` is that it will likely contain a version, but can contain a lot more information in it, including characters which break the json output in the first place (hence the main purpose of this PR). Saying 'the driver has already done a string conversion' isn't necessarily true, since it could *not* print the version in the `driverInfo`.

That said, `driverID` is significantly more robust than `vendorID`, so definitely should be preferred when available.

mikes-lunarg: Ok I can get on board with this. The new implementation opts new drivers out of the legacy vendor matching if they provide `driverID`. A nice side-effect is a more accurate profile name on my NVIDIA device!

mikes-lunarg (approved):

vulkaninfo/outputprinter.h
charles-lunarg: Yep, much more 'robust'. I also made sure to only apply the 'formatted' driverVersion to the text/html/vkconfig output, as the profile json schema requires the field be an int, not a string.

I kinda feel bad now for thinking all driverVersion fields used the Vulkan API version format, since that is patently not the case!

Issue #592, opened 2021-11-23
[cube] cross compile fails silently


SupervisedThinking:
Well is there any reason for [this](https://github.com/KhronosGroup/Vulkan-Tools/blob/master/cube/CMakeLists.txt#L231):
```
    if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL ${CMAKE_HOST_SYSTEM_PROCESSOR})
        add_executable(vkcube
                       cube.c
                       ${PROJECT_SOURCE_DIR}/cube/cube.vert
                       ${PROJECT_SOURCE_DIR}/cube/cube.frag
                       cube.vert.inc
                       cube.frag.inc
                       ${OPTIONAL_WAYLAND_DATA_FILES})
        target_link_libraries(vkcube Vulkan::Vulkan)
        include(CheckLibraryExists)
        CHECK_LIBRARY_EXISTS("rt" clock_gettime "" NEED_RT)
        if (NEED_RT)
            target_link_libraries(vkcube rt)
        endif()
    endif()
```
Because cross compiling `vkcube/vkcubepp` just fails & it fails silent. Without these checks I can cross compile vkcube on my x86_64 system e.g. vor ARM RK3399 or RPi4 - ofc the binaries will only run on those ARM target systems but this should be obvious if you cross compile.

PR #704, opened 2022-11-09
build: Compile vkcube and other demos, even when cross-compiling

smcv:
When building software for a non-x86 processor on an x86 build machine, or building 32-bit software on a 64-bit build machine, it's still useful to compile demo programs that can be copied into the target environment and run there, even if they won't necessarily be runnable on the build machine.

This also addresses a build failure when cross-compiling and attempting to leave the demos enabled:

    CMake Error at cube/CMakeLists.txt:274 (install):
      install TARGETS given target "vkcube" which does not exist.

(and similar for the other demos)

Resolves: https://github.com/KhronosGroup/Vulkan-Tools/issues/592

### DIFF
--- a/cube/CMakeLists.txt
+++ b/cube/CMakeLists.txt
@@ -228,21 +228,19 @@ include_directories(${CUBE_INCLUDE_DIRS})
 if(APPLE)
     include(macOS/cube/cube.cmake)
 elseif(NOT WIN32)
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
Yep, seems a bit silly to not allow cross compilation. 
Thanks for doing the PR!

PR #707, opened 2022-11-17
build: Update to header 1.3.235

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):
Were there any difficulties with code gen in this repo? It looks like everything was generated just fine.

mikes-lunarg:
> Were there any difficulties with code gen in this repo? It looks like everything was generated just fine.

No issues. As usual, all the harry bits ended up in VVL, mostly caused by new unions in the API.

PR #709, opened 2022-11-28
icd: Set MockICD driver version to header version

spencer-lunarg:
Currently the MockICD only exposed Vulkan 1.1 but in theory supports what ever version the spec is


charles-lunarg (approved):
Changes are fine, as it makes sense to allow mock icd to be “whatever” the current header version it is. Though in the future it would be nice to set the version dynamically somehow at runtime

Issue #711, opened 2022-11-30
consider upgrading to c++17


jeremyg-lunarg:
In https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4514 it was suggested that we move all ecosystem components to a newer c++ standard. Is this feasible for Vulkan-Tools?

charles-lunarg:
Feasible? Yes, very much. Mainly involves setting the standard to something newer and fixing any issues that comes up.

As far as 'cleaning up the code to *use* C++17', that wont happen anytime soon since it isn't a high priority.

jeremyg-lunarg:
The first part is all that is desired, I think.  The OpenBSD maintainer ran into some obscure issues in VVL that were solved by c++17, and using the same standard across the ecosystem makes life simpler for everyone.

PR #712, opened 2022-11-30
Update to cpp17

charles-lunarg:
Resolves issue #711 

Replaced the adhoc vulkaninfo_optional with std::optional, I'm sure there are plenty of other cleanup's that can be done with C++17, but this is the most egregious.

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -1,7 +1,7 @@
 /*
- * Copyright (c) 2015-2021 The Khronos Group Inc.
- * Copyright (c) 2015-2021 Valve Corporation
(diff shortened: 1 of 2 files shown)


charles-lunarg:
Turns out there is quite a bit more leg work to do to get this PR in a fit state - closing the PR while I work on that.

jeremyg-lunarg:
Also this isn't urgent and is probably good to defer until after the upcoming SDK freeze

juan-lunarg:
> Replaced the adhoc vulkaninfo_optional with std::optional, I'm sure there are plenty of other cleanup's that can be done with C++17, but this is the most egregious.

I'd be careful about doing that. See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4854

PR #713, opened 2022-11-30
vulkaninfo: Fix quoting strings in json output

charles-lunarg:
When adding a value_description, the json output would incorrectly quote the PrintKeyString output, leading to invalid json being generated.


richard-lunarg (approved):
LGTM

PR #714, opened 2022-12-01
Fix negative gpu selection in vkcube & vkcubepp

charles-lunarg:
Fix the crash when trying to select a negative GPU index and the validation error stating that the apiVersion wasn't set.


ziga-lunarg (approved):
LGTM

PR #717, opened 2022-12-01
cubepp: Make sure width & height are > 0

### LINKED ISSUE #715: Wrong error message with invalid width 0
On linux running `vkcubepp --width 0` gives the error:
`Could not find both graphics and present queues`

Which is not true. `vkcube --width 0` prints help which seems correct behaviour.

charles-lunarg:
This check was present in cube but wasn't added to cubepp.

Fixes #715

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -4075,13 +4075,27 @@ static void demo_init(struct demo *demo, int argc, char **argv) {
             i++;
             continue;
         }
-        if (strcmp(argv[i], "--width") == 0 && i < argc - 1 && sscanf(argv[i + 1], "%d", &demo->width) == 1 && demo->width > 0) {
-            i++;
-            continue;
+        if (strcmp(argv[i], "--width") == 0) {
+            if (i < argc - 1 && sscanf(argv[i + 1], "%d", &demo->width) == 1) {
+                if (demo->width > 0) {
+                    i++;
+                    continue;
+                } else {
+                    ERR_EXIT("The --width parameter must be greater than 0", "User Error");
+                }
+            }
+            ERR_EXIT("The --width parameter must be followed by a number", "User Error");
         }
-        if (strcmp(argv[i], "--height") == 0 && i < argc - 1 && sscanf(argv[i + 1], "%d", &demo->height) == 1 && demo->height > 0) {
-            i++;
-            continue;
+        if (strcmp(argv[i], "--height") == 0) {
+            if (i < argc - 1 && sscanf(argv[i + 1], "%d", &demo->height) == 1) {
+                if (demo->height > 0) {
+                    i++;
+                    continue;
+                } else {
+                    ERR_EXIT("The --height parameter must be greater than 0", "User Error");
+                }
+            }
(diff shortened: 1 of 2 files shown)


ziga-lunarg (approved):
LGTM

lunarpapillo (changes requested):
These changes do nothing to prevent `height` and `width` from being set to 0.  They do mean that if you explicitly pass `--width 0` or `--height 0` on the command line, the argument won't be recognized as part of the `--width` or `--height` switch, and will continue to be parsed (probably causing cryptic argument errors).

charles-lunarg:
> These changes do nothing to prevent `height` and `width` from being set to 0. They do mean that if you explicitly pass `--width 0` or `--height 0` on the command line, the argument won't be recognized as part of the `--width` or `--height` switch, and will continue to be parsed (probably causing cryptic argument errors).

This change brings the width/height parameter into alignment with the behavior of vkcube & the gpu_number argument for both vkcube and vkcubepp. I would rather 'do what we already do consistently' than anything else.

charles-lunarg:
> These changes do nothing to prevent height and width from being set to 0. 
There is no way to prevent that because this is a command line argument being passed in.
Passing in 0 to vkcube causes help to be printed. passing in 0 to vkcubepp will cause a crash cause the window size must be 1 or greater. Better to return help rather than crash.

richard-lunarg:
The .cpp version declares the width and height as unsigned integers, while the C version declares them as signed integers. Thus the CPP version's scanf will never return a negative value ;-)

cube/cube.c
```diff
@@ -4075,13 +4075,21 @@ static void demo_init(struct demo *demo, int argc, char **argv) {
...
         }
-        if (strcmp(argv[i], "--width") == 0 && i < argc - 1 && sscanf(argv[i + 1], "%d", &demo->width) == 1 && demo->width > 0) {
-            i++;
-            continue;
+        if (strcmp(argv[i], "--width") == 0 && i < argc - 1) {
+            if (sscanf(argv[i + 1], "%d", &demo->width) == 1 && demo->width > 0) {
+                i++;
+                continue;
+            } else {
+                ERR_EXIT("The --width parameter must be greater than 0", "User Error");
```
lunarpapillo, line 4083: If the user passes something like `--width max`, the error message will be somewhat misleading.  (I'm not sure this corner case is worth clarifying, however, though I do note that it's not much more code to make the error cases clear...)

lunarpapillo: Also, there's a weird corner case if the user provides `--width` as the last parameter - no error is provoked and the width is silently unchanged.  That may be more of a concern.

lunarpapillo (commented):
A couple of observations are listed.  But you should definitely consider a review from a code owner more important than anything I'd say.

cube/cube.c
charles-lunarg: I've modified the code to respect whether the user has put in a number or not. If they did and its negative, it gives the correct error. If they don't provide a number, it now prints that message.

charles-lunarg:
Bob gave in person approval (plus the issues he raised were resolved). Merging.

PR #720, opened 2022-12-03
ci: Speed up MacOS builds

juan-lunarg:
Caches update_deps.py dependencies,
In particular MoltenVK since it's a huge dependency.


charles-lunarg (approved):
Awesome!!

PR #723, opened 2022-12-14
vulkaninfo: Remove macOS application bundle

mikes-lunarg:
macOS Ventura broke the "shell script as a executable" method for delivering vulkaninfo as an .app bundle. Rather than find another workaround, we decided to remove the bundle version from the SDK since it is redundant. The command-line vulkaninfo is unaffected and can still be launched from Finder.


richard-lunarg (approved):
LGTM. Time to get rid of some of that waxy build up...

PR #724, opened 2022-12-20
build: Update to header 1.3.238

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):
Yaaaay gone are the #ifdef VK_ENABLE_BETA_EXTENSIONS eyesores!

PR #726, opened 2022-12-27
Use pkg-config to find X11 / XCB

juan-lunarg:
Consistent with VVL's approach

Reduces amount of custom CMake code we have to maintain


juan-lunarg:
I'd like to do the same for Wayland / DirectFB, but in other PRs.

cube/CMakeLists.txt
```diff
@@ -49,19 +49,20 @@ else()
...
     option(BUILD_WSI_XLIB_SUPPORT "Build Xlib WSI support" ON)
     option(BUILD_WSI_WAYLAND_SUPPORT "Build Wayland WSI support" ON)
     option(BUILD_WSI_DIRECTFB_SUPPORT "Build DirectFB WSI support" OFF)
     set(CUBE_WSI_SELECTION "XCB" CACHE STRING "Select WSI target for vkcube (XCB, XLIB, WAYLAND, DIRECTFB, DISPLAY)")
 
+    find_package(PkgConfig REQUIRED QUIET) # Use PkgConfig to find Linux system libraries
+
     if(BUILD_WSI_XCB_SUPPORT)
-        find_package(XCB REQUIRED)
+        pkg_check_modules(XCB REQUIRED QUIET IMPORTED_TARGET xcb)
```
charles-lunarg, line 61: I assume that pkg_check_modules is akin to find_package, in that if REQUIRED is specified and the package isn't found, it'll abort?

juan-lunarg: Correct.

charles-lunarg (approved):

PR #727, opened 2022-12-27
cmake: Use PkgConfig to find DirectFB

juan-lunarg:


charles-lunarg (approved):
Changes look good.

I did want to give a heads up to @caramelli that this change is happening, not that the changes should affect anything.

PR #731, opened 2023-01-13
vkcube: Add more debug utils object names and cb labels

jeremyg-lunarg:


cube/cube.c
```diff
@@ -2002,6 +2034,7 @@ static VkShaderModule demo_prepare_shader_module(struct demo *demo, const uint32
 
     err = vkCreateShaderModule(demo->device, &moduleCreateInfo, NULL, &module);
     assert(!err);
+    demo_name_object(demo, VK_OBJECT_TYPE_SHADER_MODULE, (uint64_t)module, "%s", name);
```
charles-lunarg, line 2037: Couldn't this be simplified to pass name as the string directly? For example:
```c
    demo_name_object(demo, VK_OBJECT_TYPE_SHADER_MODULE, (uint64_t)module, name);
```

charles-lunarg (commented):
Just a general comment, but its a good idea to check the error code returned from `SetDebugUtilsObjectNameEXT`, as well as make sure any other functions that return VkResult are being checked. I found `vkGetPhysicalDeviceSurfaceSupportKHR` (line 3714) isn't being checked, and I'm sure there are others. Not a requirement, just a thought.

cube/cube.c
jeremyg-lunarg: Some code checking tools become angry when you pass a non-literal string as the format argument to a printf-like function.

charles-lunarg: Ahhh okay, I figured that might of been the case. Good to know for sure.

PR #735, opened 2023-01-27
v1.3.240 update

mikes-lunarg:
Normal header stuff + same GN updates from Vulkan-Loader


charles-lunarg (approved):
Appreciate you handling all of these various GN build issues :D

PR #740, opened 2023-02-15
mockicd: Add more PhysicalDevice functions not in profiles

spencer-lunarg:
There are many `vkEnumeratePhysicalDevice*` and `vkGetPhysicalDevice` function that just giving an arbitrary value provides more test coverage when using MockICD with Profiles to test Validation


charles-lunarg (approved):
Everything here looks 'normal'.
Its useful for these functions to return real values rather than nothing (and fixes some issues where VVL expects values but the driver never writes them).

PR #744, opened 2023-02-16
build: Update to header 1.3.241

mikes-lunarg:
- Update known-good
- Tweak generate_vulkan_wrapper.py to ignore sci header
- Generate source


mikes-lunarg:
This won't pass CI until the Vulkan-Loader tag is pushed

mikes-lunarg:
This should be ready to go now

scripts/common_codegen.py
```diff
@@ -60,6 +60,7 @@
     'provisional' : 'VK_ENABLE_BETA_EXTENSIONS',
     'directfb' : 'VK_USE_PLATFORM_DIRECTFB_EXT',
     'screen' : 'VK_USE_PLATFORM_SCREEN_QNX',
+    'sci' : 'VK_USE_PLATFORM_SCI',
```
charles-lunarg, line 63: This should be taken from the spec... but not going to hold up the PR for it.

charles-lunarg (approved):

PR #745, opened 2023-02-19
mockicd: Add more Memory Types

spencer-lunarg:
Expands MockICD to have more Memory Types found on various implementations


charles-lunarg (approved):
👍

PR #759, opened 2023-03-22
build: Update to header 1.3.244

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (changes requested):
The memory funcs look to be missing useful bodies.

mikes-lunarg:
> The memory funcs look to be missing useful bodies.

Good catch! The mock has to actually pass a valid pointer back to the application or else things will get a bit crashy.

charles-lunarg (approved):
Great!

PR #760, opened 2023-03-24
vulkaninfo: Dont enable Direct Driver Loading Ext

charles-lunarg:
The VK_LUNARG_direct_driver_loading extension, when enabled, causes spurious warnings in the loader. Vulkaninfo should just not enable it to prevent these warnings from apperaing in the output.

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -652,7 +652,10 @@ struct AppInstance {
         AppCompileInstanceExtensionsToEnable();
 
         std::vector<const char *> inst_exts;
-        for (const auto &ext : inst_extensions) inst_exts.push_back(ext.c_str());
+        for (const auto &ext : inst_extensions) {
+            if (ext == "VK_LUNARG_direct_driver_loading") continue;  // skip this extension since it triggers warnings in the loader
+            inst_exts.push_back(ext.c_str());
+        }
 
         const VkInstanceCreateInfo inst_info = {
             VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO,


charles-lunarg:
The warning that is avoided with this PR
```
WARNING: [Loader Message] Code 0 : loader_scan_for_direct_drivers: The VK_LUNARG_direct_driver_loading extension was enabled but the pNext chain of VkInstanceCreateInfo did not contain the VkDirectDriverLoadingListLUNARG structure.
```

mikes-lunarg (approved):
LGTM. I'm sorta curious why the warning exists in the first place, enabling and not using the extension should be harmless. This feels more like a best practices check.

charles-lunarg:
Because `VK_LUNARG_direct_driver_loading` pretty much takes place before layer loading, so if the user intends to use this extension but fails to add the pNext chain, (and there are no drivers on the system), the instance will fail to be created before any best practices checks can run.

juan-lunarg (approved):

juan-lunarg:
Oh I see the problem this causes. I used `vulkaninfo --summary | grep "Vulkan Instance Version"` as part of my `.bash_aliases` as a sanity check to know which version of the SDK I'm using.

```
WARNING: [Loader Message] Code 0 : loader_scan_for_direct_drivers: The VK_LUNARG_direct_driver_loading extension was enabled but the pNext chain of VkInstanceCreateInfo did not contain the VkDirectDriverLoadingListLUNARG structure.
Vulkan Instance Version: 1.3.243
```

Is this going to be cherry picked?

charles-lunarg:
Yes it should be. I can do that in a bit but feel free to do it for me if you are able.

PR #762, opened 2023-03-27
vkcubepp: Fix custom height not working

charles-lunarg:


cube/cube.cpp
```diff
@@ -906,7 +906,7 @@ void Demo::init(int argc, char **argv) {
         if (strcmp(argv[i], "--height") == 0) {
             int32_t in_height = 0;
             if (i < argc - 1 && sscanf(argv[i + 1], "%d", &height) == 1) {
-                if (height > 0) {
+                if (in_height > 0) {
```
ziga-lunarg, line 909: Shouldn't this still be `height` and in the line below in_height = height?

ziga-lunarg (approved):

PR #766, opened 2023-04-05
vulkaninfo: Add Android build support

andy-poole-samsung:
This change adds a build step to "build-android/build_all.sh" to create a command-line version of `vulkaninfo` for Android devices. Once built, `vulkaninfo` can be pushed to the device and executed using `adb`.

Because it is not possible to create `ANativeWindow` from a native command-line application, `VULKANINFO_WSI_ENABLED` is not defined for Android builds of `vulkaninfo`.

### DIFF
--- a/build-android/build_all.sh
+++ b/build-android/build_all.sh
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+ANDROID_BUILD_DIR=$(dirname $(readlink -f $0))
+PROJECT_DIR=$ANDROID_BUILD_DIR/..
+
 if [ -z "${ANDROID_SDK_HOME}" ];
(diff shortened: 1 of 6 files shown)


charles-lunarg (approved):
Thanks for the contribution!

Code looks correct, but since android wasn't supported for a long time its not something that can be tested in our internal CI. I will still push it through to make sure other platforms dont break but I assume you've checked that it works locally.

andy-poole-samsung:
@charles-lunarg no problem at all - I am glad to contribute this. Upstreaming this change has been in my TODO list for a while now! Thank you for approving and pushing it through CI (not sure what's happened to the queued checks)

charles-lunarg:
So I'm going to merge this because the failed/delayed checks are Ubuntu 18 jobs, which has just been removed from the github actions. Thats gonna have to be fixed pretty soon but thankfully all the other jobs succeeded.

PR #767, opened 2023-04-05
vulkaninfo: Add "VK_" prefix for PrintFlagBits

andy-poole-samsung:
For types like VkSampleCountFlagBits a "VK_" prefix will be added for JSON outputs. This makes the behaviour consistent for JSON output.

Fields that currently use a type of `VkSampleCountFlags` will use a `VK_` prefix in the JSON output of `vulkaninfo`. However, fields that use a type of `VkSampleCountFlagBits` will not have this prefix. For example:

VkSampleCountFlags:
```json
"VkPhysicalDeviceSampleLocationsPropertiesEXT": {
    "sampleLocationSampleCounts": [
        "VK_SAMPLE_COUNT_1_BIT",
        "VK_SAMPLE_COUNT_2_BIT",
        "VK_SAMPLE_COUNT_4_BIT"
    ],
},
```
VkSampleCountFlagBits:
```json
"VkPhysicalDeviceFragmentShadingRatePropertiesKHR": {
    "maxFragmentShadingRateRasterizationSamples": "SAMPLE_COUNT_4_BIT",
},
```
The above inconsistent behaviour makes it difficult to post-process JSON output from `vulkaninfo`. This change adds the `VK_` prefix to fields that use a `FlagBits` type to ensure consistent behaviour.

### DIFF
--- a/scripts/vulkaninfo_generator.py
+++ b/scripts/vulkaninfo_generator.py
@@ -475,8 +475,12 @@ def PrintFlags(bitmask, name):
 def PrintFlagBits(bitmask):
     return f"""void Dump{bitmask.name}(Printer &p, std::string name, {bitmask.name} value) {{
     auto strings = {bitmask.name}GetStrings(value);
-    if (strings.size() > 0)
-        p.PrintKeyString(name, strings.at(0));
+    if (strings.size() > 0) {{
+        if (p.Type() == OutputType::json)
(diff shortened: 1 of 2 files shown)


charles-lunarg (approved):
Again thanks for the fix!

Seems like an oversight on my part. Vulkaninfo is not well tested, and apparently the Schema for profiles which this was checked against doesn't check the naming convention of things.

PR #768, opened 2023-04-06
ci: Use Ubuntu 22 instead of 18

charles-lunarg:


mikes-lunarg (approved):
LGTM, the proof is in the all of those green check marks!

PR #772, opened 2023-04-14
vulkaninfo: Flush before exiting

charles-lunarg:
Vulkaninfo does not manually flush the output before exiting, which means that if a crash were to happen during teardown, the stdout buffer may not have finished writing.

To implement this, it was necessary to turn the destructor of Printer into a function, as the lifetime of the printer is longer than the various Vulkan objects that are liable to crash.

This PR comes about to work around issues where some layers/drivers/systems will crash during vkDestroyDevice.

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -1139,6 +1139,8 @@ int main(int argc, char **argv) {
 
         RunPrinter(*(printer.get()), parse_data, instance, gpus, surfaces);
 
+        // Call the printer's destructor before the file handle gets closed
+        printer.reset(nullptr);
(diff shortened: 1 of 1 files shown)


jeremyg-lunarg (approved):
LGTM

vulkaninfo/vulkaninfo.cpp
```diff
@@ -1139,6 +1139,8 @@ int main(int argc, char **argv) {
 
         RunPrinter(*(printer.get()), parse_data, instance, gpus, surfaces);
 
+        printer->FinishOutput();
```
jeremyg-lunarg, line 1142: It seems like you could have also called `printer.reset();` here instead.

jeremyg-lunarg (commented):
wow that is a lot less code change

PR #780, opened 2023-04-27
mock: Add support for vkGetDeviceImageMemoryRequirements

juan-lunarg:
Needed for KhronosGroup/Vulkan-ValidationLayers/issues/5616


charles-lunarg (approved):
Changes look good - just needs to pass CI (aka - fix the autogen check)

PR #786, opened 2023-05-06
mock: Extend external memory functionality

artem-lunarg:
`GetPhysicalDeviceExternalBufferProperties` changes:
Previous implementation violated property that handle type is at least compatible with itself.
It's needed to implement testing for VUID-VkMemoryAllocateInfo-pNext-00639

`GetFenceWin32HandleKHR/GetFenceFdKHR` changes:
Return not null handles to pass VUID-VkImportFenceWin32HandleInfoKHR-handle-01462

Tested with GHA using this fork: https://github.com/artem-lunarg/Vulkan-ValidationLayers/tree/external-memory2


charles-lunarg (approved):
Changes look appropriate for what is needed. Writing default values when appropriate is good, as well as writing non zero handles.

PR #790, opened 2023-05-16
mock: Fix sparse image 2 gets

scottmarsland:
Fix vkGetPhysicalDeviceSparseImageFormatProperties2 and
vkGetPhysicalDeviceSparseImageFormatProperties2KHR queries when
pProperties is null.
Fix vkGetImageSparseMemoryRequirements2 and
vkGetImageSparseMemoryRequirements2KHR queries when
pSparseMemoryRequirements is null.


charles-lunarg (approved):
Makes sense - GetPhysicalDeviceSparseImageFormatProperties2KHR and GetImageSparseMemoryRequirements are defined to only return one element, so pointing directly to that is fine.

PR #792, opened 2023-05-22
mock: Parameterization changes for Vulkan SC

aqnuep:


juan-lunarg (approved):
Changes look good to me. But I'd appreciate another pair of eyes @charles-lunarg / @spencer-lunarg

spencer-lunarg (approved):
this is nice because I would have not known `VK_KHR_object_refresh` was something added to SC, was confused when I saw it right now at first

PR #793, opened 2023-05-22
vulkaninfo: Reduce NVIDIA quaternary version field to 6 bits

MarijnS95:
The tertiary version field is shifted 6 bits to the right, meaning there are only 6 bits left to for the quaternary version field, not 10. See also:

https://github.com/SaschaWillems/vulkan.gpuinfo.org/blob/1e6ca6e3c0763daabd6a101b860ab4354a07f5d3/functions.php#L305

CC @SaschaWillems 

(EDIT: I did not test or run into this problem, just noticed this oddity while trying to understand other version parsing quirks)


charles-lunarg (approved):
Whoops! This seems like a silly oversight, possibly due to a copy paste error.

MarijnS95:
@charles-lunarg I should have done so but maybe it was on another repo? We're a Khronos member anyway so I'm not sure if the CLA still makes sense then, but let me check if the commit was made under my personal email..

charles-lunarg:
Ah okay great, seems to have worked now. I'll merge since this passed CI before.

MarijnS95:
I signed it manually again, clicking `recheck` after pushing the patch under my @Traverse-Research email didn't change anything :)

PR #797, opened 2023-05-24
Update known_good.json for MoltenVK 1.2.4

richard-lunarg:
Updated known good for new MoltenVK release.

### DIFF
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -14,7 +14,7 @@
       "sub_dir" : "MoltenVK",
       "build_dir" : "MoltenVK",
       "install_dir" : "MoltenVK",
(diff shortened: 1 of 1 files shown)


scripts/known_good.json
```diff
@@ -14,7 +14,7 @@
       "sub_dir" : "MoltenVK",
       "build_dir" : "MoltenVK",
       "install_dir" : "MoltenVK",
-      "commit" : "v1.2.3",
+      "commit" : "v1.2.4",
```
lunarpapillo, line 17: Technically, this is a `tag`, not a `commit`... but it's no worse than its predecessor.

lunarpapillo (approved):

PR #800, opened 2023-05-25
Fix cmd buffer/pool leak in vkcubepp

### LINKED ISSUE #795: vkcubepp leaks a command buffer and a command pool if iconized before stopping
From 1.3.250 SDK Testing:
Using SDK 1.3.250, Win 10, AMD, from powershell window run vkcubepp --validate.  Iconify the cubepp window and restore it.  Hit ESC to end the program and there are two popups complaining about a command buffer and a command pool that haven't been destroyed.
This doesn't happen with vkcube.  This doesn't happen if you don't iconify vkcubepp.

Tony-LunarG:
Fixes #795

### DIFF
--- a/cube/cube.cpp
+++ b/cube/cube.cpp
@@ -1553,13 +1553,12 @@ void Demo::init_vk_swapchain() {
 }
 
 void Demo::prepare() {
-    prepare_init_cmd();
-
     prepare_buffers();
     if (is_minimized) {
         prepared = false;
         return;
     }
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
Had to look at the rest of the source code, but this appears to work correctly.

prepare_buffers really should be 'prepare swapchain' but I digress. It contains an early exit based on 'is_minimized' so it works. Heck, prepare_buffers is what updates is_minimized.

PR #802, opened 2023-05-26
vulkaninfo: Only enable necessary instance extensions

### LINKED ISSUE #799: vulkaninfo enables allllllll instance extensions which provokes validation errors
Certain instance extensions have conditions for their enablement that vulkaninfo blatantly ignores, triggering validation errors.

The solution is to have vulkaninfo only enable the required extensions. These should be:
* debug_report
* surface
*<platform_specific_surface_extension>
* maybe another one I am forgetting or not aware of.

This issue was originally found during [SDK testing](https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/-/issues/1299).

charles-lunarg:
Previously, vulkaninfo would enable all instance extensions, which caused validation errors to be generated. Now, vulkaninfo only enables extensions it knows it needs to, such as:
VK_EXT_debug_report
VK_KHR_get_physical_device_properties2
VK_KHR_portability_enumeration
VK_KHR_surface
<platform specific surface extensions>
VK_KHR_get_surface_capabilities2
VK_KHR_surface_protected_capabilities

Fixes #799

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -653,7 +653,6 @@ struct AppInstance {
 
         std::vector<const char *> inst_exts;
         for (const auto &ext : inst_extensions) {
-            if (ext == "VK_LUNARG_direct_driver_loading") continue;  // skip this extension since it triggers warnings in the loader
             inst_exts.push_back(ext.c_str());
         }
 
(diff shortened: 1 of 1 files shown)


vulkaninfo/vulkaninfo.h
```diff
@@ -715,9 +714,90 @@ struct AppInstance {
         global_extensions = AppGetGlobalLayerExtensions(nullptr);
     }
     void AppCompileInstanceExtensionsToEnable() {
-        // Get all supported Instance extensions (excl. layer-provided ones)
         for (const auto &ext : global_extensions) {
-            inst_extensions.push_back(ext.extensionName);
+            if (strcmp(VK_EXT_DEBUG_REPORT_EXTENSION_NAME, ext.extensionName) == 0) {
```
mikes-lunarg, line 718: I have a slight preference for putting the extension names in an array, up to you though.

mikes-lunarg (approved):
LGTM

vulkaninfo/vulkaninfo.h
charles-lunarg: Array, like a std::array? Not sure how to handle dynamicism using a std::array.

mikes-lunarg: Yeah I suppose it would have to be something like a vector that can handle #ifdefs in the initializer list. Or you could a braced init list directly in a range based for loop. Maybe more trouble than it is worth.

PR #808, opened 2023-06-02
scripts/CMakeLists.txt: Preserve CMAKE_FIND_ROOT_PATH passed in by user

kanavin:
…

Resetting CMAKE_FIND_ROOT_PATH in particular breaks builds in Yocto (which is a major cross compiling framework). If someone is cross compiling, they typically configure cmake itself to look in the right places.


scripts/CMakeLists.txt
```diff
@@ -112,9 +112,3 @@ endif()
 if (MOLTENVK_REPO_ROOT)
     list(APPEND CMAKE_PREFIX_PATH ${MOLTENVK_REPO_ROOT})
 endif()
-
-if (CMAKE_CROSSCOMPILING)
-    set(CMAKE_FIND_ROOT_PATH ${CMAKE_PREFIX_PATH} PARENT_SCOPE)
```
juan-lunarg, line 117: I've been working on adding Android to our builds in the validation layers I added this logic to simplify finding dependencies on Android since CMAKE_PREFIX_PATH was getting ignored.

I'm open to a more idiomatic solution.

scripts/CMakeLists.txt
```diff
@@ -112,9 +112,3 @@ endif()
...
-else()
-    set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} PARENT_SCOPE)
```
juan-lunarg, line 119: Removing this will break our CI.

scripts/CMakeLists.txt
kanavin: I think the right way to do this is to have CI write 
```
set(CMAKE_FIND_ROOT_PATH ${CMAKE_PREFIX_PATH} PARENT_SCOPE)
```
into an external file, e.g. android-toolchain.cmake, then run cmake with
```
-DCMAKE_TOOLCHAIN_FILE=android-toolchain.cmake
```

https://cmake.org/cmake/help/latest/variable/CMAKE_TOOLCHAIN_FILE.html

That way you won't step on other environments that set CMAKE_FIND_ROOT_PATH in a way appropriate for them.

juan-lunarg: We can't modify the android toolchain. Please see how we build for Android.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/3c3365f47748fbf08970fda9ed26ec0a29a8f44d/.github/workflows/vvl.yml#L175

kanavin: Actually, appending instead of replacing will work just as well:
```
set(CMAKE_FIND_ROOT_PATH ${CMAKE_FIND_ROOT_PATH} ${CMAKE_PREFIX_PATH} PARENT_SCOPE)
```
Is that ok for you?

juan-lunarg: > set(CMAKE_FIND_ROOT_PATH ${CMAKE_FIND_ROOT_PATH} ${CMAKE_PREFIX_PATH} PARENT_SCOPE)

That solution seems good to me.

kanavin: Alright, I pushed a different commit to this PR that does that.

juan-lunarg (approved):

juan-lunarg:
Thanks for the fix @kanavin 👍🏾

PR #810, opened 2023-06-12
build: Update to header 1.3.253

mikes-lunarg:
- Update known-good
- Generate source


common/vulkan_wrapper.cpp
```diff
@@ -579,6 +579,11 @@ int InitVulkan(void) {
 #endif
 
 #ifdef VK_ENABLE_BETA_EXTENSIONS
+    vkGetPhysicalDeviceVideoEncodeQualityLevelPropertiesKHR =
+        reinterpret_cast<PFN_vkGetPhysicalDeviceVideoEncodeQualityLevelPropertiesKHR>(
+            dlsym(libvulkan, "vkGetPhysicalDeviceVideoEncodeQualityLevelPropertiesKHR"));
+    vkGetEncodedVideoSessionParametersKHR =
+        reinterpret_cast<PFN_vkGetEncodedVideoSessionParametersKHR>(dlsym(libvulkan, "vkGetEncodedVideoSessionParametersKHR"));
```
charles-lunarg, line 586: this looks funky - why are we dlsym'ing extension functions? (or has this always been like this and I never noticed)

mikes-lunarg:
> CI Vulkan-Tools build # 1023 failed.

Oh good, it's not just my machine. Something is going wrong with Vulkan-Hpp:

```
In file included from /home/mike/Projects/HeaderUpdate/build/Vulkan-Headers/build/install/include/vulkan/vulkan.hpp:6794,
                 from /home/mike/Projects/HeaderUpdate/build/Vulkan-Tools/cube/cube.cpp:43:
/home/mike/Projects/HeaderUpdate/build/Vulkan-Headers/build/install/include/vulkan/vulkan_handles.hpp:12039:114: error: ‘StructureChainAllocator’ does not name a type; did you mean ‘StructureChainValidation’?
12039 |               typename B2                                                                                      = StructureChainAllocator,
      |                                                                                                                  ^~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                                                  StructureChainValidation
/home/mike/Projects/HeaderUpdate/build/Vulkan-Headers/build/install/include/vulkan/vulkan_handles.hpp:12040:91: error: type/value mismatch at argument 2 in template parameter list for ‘template<class, class> struct std::is_same’
12040 |               typename std::enable_if<std::is_same<typename B2::value_type, StructureChain>::value, int>::type = 0>
      |                                                                                           ^
/home/mike/Projects/HeaderUpdate/build/Vulkan-Headers/build/install/include/vulkan/vulkan_handles.hpp:12040:91: note:   expected a type, got ‘StructureChain’
/home/mike/Projects/HeaderUpdate/build/Vulkan-Headers/build/install/include/vulkan/vulkan_handles.hpp:12040:104: error: template argument 1 is invalid
12040 |               typename std::enable_if<std::is_same<typename B2::value_type, StructureChain>::value, int>::type = 0>
      |                                                                                                        ^
/home/mike/Projects/HeaderUpdate/build/Vulkan-Headers/build/install/include/vulkan/vulkan_handles.hpp:12040:105: error: ‘<declaration error>’ is not a template [-fpermissive]
12040 |               typename std::enable_if<std::is_same<typename B2::value_type, StructureChain>::value, int>::type = 0>
      |                                                                                                         ^~
/home/mike/Projects/HeaderUpdate/build/Vulkan-Headers/build/install/include/vulkan/vulkan_handles.hpp:12043:44: error: ‘StructureChainAllocator’ has not been declared
12043 |                                            StructureChainAllocator &                                            structureChainAllocator,
      |                                            ^~~~~~~~~~~~~~~~~~~~~~~
```

charles-lunarg (commented):
Seems we will have to get a fix for vulkan-hpp before we can get this going.

common/vulkan_wrapper.cpp
mikes-lunarg: I believe it has always been this way. vulkan_wrapper is an entry point loader that came from the Android folks, but I'm not sure who actually uses it.

charles-lunarg: Ah, well I'll resolve this comment then. Seems very suspicious. This would return nullptr for any desktop users of this code. Still, if its from android, then I can't claim its "wrong". (they do have their own loader after all)

mikes-lunarg:
Succeeded by #815

PR #813, opened 2023-06-15
icd: Move handwritten code into .cpp file

charles-lunarg:
Takes the hand written portions of the mock_icd_generator.py and puts them in a .cpp file, then moves the code gen into function_declarations.inl and function_definitions.inl.

This move makes it easier to maintain the C++ code that exists because it no longer lives in a python file. This commit does not make adding custom code easier, as it still is in the python file. The intent of this commit is to create a baseline that would allow such code to be put inside of C++ instead of python.

Additionally, this commit removes egregious over engineering in the mock icd CMakeLists.txt. Things removed are macros that are only called once, foreach loops over single items, and replacing variables that had no solid reason to be a variable.

This PR is as much a request for comments as it is a real PR. I am not 100% on the approach used here, and in all likelyhood this PR should not be merged until there are tests to validate I didn't break anything basic first. (since this repo has essentially zero tests).


juan-lunarg (commented):
Overall the changes look good. I'll see if I can pull down locally and test everything to make sure it's all good for vvl.

spencer-lunarg (approved):
I think this is a step in the right direction, happy to look at this as an initial clean PR

I will try to find time today (or soon) to make sure this didn't break anything in VVL (if Juan doesn't beat me to it)

charles-lunarg:
I fixed the build issues, as well as decided to make the .inl files into proper header files. That should allow code completion to work on them much more easily.

Which is important because I had a wonderfully evil idea - when running the codegen, load the existing function_definitions.h, grab the 'implementation' out of each function, then put that into the body of each function! This would allow easy modification of the code *without* having to modify the python codegen. Downside is that it makes codegen not idempotent as it will use its own output as an input... Feel free to discourage me from this dark path.

juan-lunarg (approved):
Ran mock tests locally and it passed fine 👍🏾

juan-lunarg:
> Ran mock tests locally and it passed fine 👍🏾

If you want to be extra sure everything is fine I'd modify the known_good.json in VVL and test the PR in a fork making sure CI passes.

charles-lunarg:
FYI this PR will very likely require rebasing ontop of #813, so I will hold off further updates until that lands (as well as once that lands I will use it to write more tests for MockICD!)

juan-lunarg:
> FYI this PR will very likely require rebasing ontop of #813, so I will hold off further updates until that lands (as well as once that lands I will use it to write more tests for MockICD!)

This is #813?

charles-lunarg:
:facepalm: Meant to say #814 

It'd be interesting to try to wait for the PR to merge before merging the PR...

PR #814, opened 2023-06-15
Add test infrastructure

charles-lunarg:
Add the necessary structure to the codebase to enable test framework code to be written along with actual tests.

This PR contains 1 test - mostly to showcase the rough outline of how tests will be done.


mikes-lunarg (approved):
```
Test project /home/mike/Projects/Vulkan-Tools/build
    Start 1: MockICD.Basic
1/1 Test #1: MockICD.Basic ....................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.01 sec
```
:1st_place_medal:

tests/main.cpp
```diff
@@ -0,0 +1,46 @@
...
+#if !defined(NDEBUG)
+    _CrtSetReportMode(_CRT_WARN, _CRTDBG_MODE_FILE);
+    _CrtSetReportFile(_CRT_ASSERT, _CRTDBG_FILE_STDERR);
+#endif
+    // Avoid "Abort, Retry, Ignore" dialog boxes
+    _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
+    SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX);
+    _CrtSetReportMode(_CRT_ASSERT, _CRTDBG_MODE_FILE);
+    _CrtSetReportFile(_CRT_ASSERT, _CRTDBG_FILE_STDERR);
+#endif
```
juan-lunarg, line 39: Do we need this?

charles-lunarg: This was taken from VVL, and this code is to prevent popups from appearing which hang the machine until someone logs in and clicks it away. Very much kept as a 'better safe than sorry' type measure.

juan-lunarg: That's why I want us to use ctest for Windows.

> Prior to CMake 3.11, interactive mode on Windows allowed system debug popup windows to appear. Now, due to CTest's use of libuv to launch test processes, all system debug popup windows are always blocked.

charles-lunarg: Ooooo I didn't know about that! So I will remove this since I expect to use ctest to launch the tests in a CI environment. Locally, I may run the executable manually but then its my own problem, not CI's.

PR #815, opened 2023-06-16
build: Update to header 1.3.254

mikes-lunarg:
- Update known-good
- Generate source
- Only build vulkaninfo and mockicd with -DVK_ENABLE_BETA_EXTENSIONS instead of the entire repo


mikes-lunarg:
Since we are still working through Vulkan-Hpp problems with VK_ENABLE_BETA_EXTENSIONS, I tried an alternate solution of turning off beta extensions for cube and and cubepp.

CMakeLists.txt
```diff
@@ -25,9 +25,6 @@ add_subdirectory(scripts)
 # This section contains variables that affect development GUIs (e.g. CMake GUI and IDEs), such as option(), folders, and variables
 # with the CACHE property.
 
-# Enable beta Vulkan extensions
-add_definitions(-DVK_ENABLE_BETA_EXTENSIONS)
```
charles-lunarg, line 29: Should there be a reminder to re-enable it when the vulkan-hpp issues are resolved? I see that its specifically enabled for vulkaninfo *only*, which is fine by me. (Which assumes nothing else wants beta extensions enabled).

charles-lunarg (approved):
Small nit: otherwise looks okay.

CMakeLists.txt
mikes-lunarg: Aside from fixing this particular issue, I think removing the global VK_ENABLE_BETA_EXTENSIONS is actually better long term. Individual targets like vulkaninfo and mockicd that want all that beta fun can opt in because they actually use it.

PR #821, opened 2023-06-20
docs: Remove unneccessary info about ccache and compile commands

juan-lunarg:


juan-lunarg:
Mac build took 3 minutes to run.

PR #823, opened 2023-06-22
icd: Document parametization for VulkanSC

charles-lunarg:
Added so that future maintainers know why these variables exist.

### DIFF
--- a/icd/CMakeLists.txt
+++ b/icd/CMakeLists.txt
@@ -16,6 +16,8 @@
 # limitations under the License.
 # ~~~
 
+# These variables enable downstream users to customize the CMake targets
+# based on the target API variant (e.g. Vulkan SC)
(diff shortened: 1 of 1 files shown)


icd/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@
 # limitations under the License.
 # ~~~
 
+# Parametization is used for VulkanSC builds
```
juan-lunarg, line 19: I'd prefer this match the existing comment in VVL

```suggestion
# These variables enable downstream users to customize the CMake targets
# based on the target API variant (e.g. Vulkan SC)
```

juan-lunarg (changes requested):

juan-lunarg (approved):

PR #824, opened 2023-06-22
Expand mock icd tests and add Address Sanitizer to github actions runs

charles-lunarg:


.github/workflows/tools.yml
```diff
@@ -76,7 +76,7 @@ jobs:
               with:
                 key: ${{ runner.os }}-${{ matrix.config }}-${{ matrix.cc }}-${{matrix.cube_wsi}}
             - name: Configure
-              run: cmake -S. -B build -G "Ninja" -DCMAKE_BUILD_TYPE=${{matrix.config}} -DCUBE_WSI_SELECTION=${{matrix.cube_wsi}} -D UPDATE_DEPS=ON -D INSTALL_ICD=ON -D BUILD_TESTS=ON
+              run: cmake -S. -B build -G "Ninja" -DCMAKE_BUILD_TYPE=${{matrix.config}} -DCUBE_WSI_SELECTION=${{matrix.cube_wsi}} -D UPDATE_DEPS=ON -D INSTALL_ICD=ON -D BUILD_TESTS=ON -D ENABLE_ADDRESS_SANITIZER=ON
```
juan-lunarg, line 79: I'd rather have address sanitizer set via the CXXFLAGS, CFLAGS, and LDFLAGS like in VVL.

tests/CMakeLists.txt
```diff
@@ -44,5 +44,10 @@ target_include_directories(vulkan_tools_tests PRIVATE ${CMAKE_CURRENT_BINARY_DIR
 target_link_libraries(vulkan_tools_tests GTest::gtest Vulkan::Vulkan)
 add_dependencies(vulkan_tools_tests generate_binary_locations)
 
+if (ENABLE_ADDRESS_SANITIZER)
+    target_compile_options(vulkan_tools_tests PUBLIC -fsanitize=address)
+    target_link_options(vulkan_tools_tests PUBLIC -fsanitize=address)
+endif ()
+
```
juan-lunarg, line 51: Currently only `vulkan_tools_tests` has the asan flags applied. What about the mock icd?

charles-lunarg: The flags are already being set for all the other targets, hence why it doesn't appear in the change list.

.github/workflows/tools.yml
charles-lunarg: Can do.

spencer-lunarg (approved):
tests look good, for now a single file seems fine (if we do start adding a lot more would be nice to split files up)

Will leave the cmake/yml suggestion to Juan

juan-lunarg (approved):

PR #826, opened 2023-06-26
cmake: Update update_deps

juan-lunarg:


scripts/update_deps.py
```diff
@@ -668,7 +671,7 @@ def main():
         '--api',
         dest='api',
         default='vulkan',
-        choices=['vulkan', 'vulkansc'],
```
charles-lunarg, line 671: Does this affect sc at all? Or does sc do it’s own modifications making this redundant?

juan-lunarg: This is taken straight from VVL. If there is an issue it would have to be fixed in vvl as well. And I use VVL as the baseline for correctness.

charles-lunarg: Works for me!

charles-lunarg (approved):

PR #828, opened 2023-06-30
build: Update to header 1.3.256

mikes-lunarg:
- Update known-good
- Generate source
- cube.cpp: Move X11 includes after vulkan.hpp because Xlib.h defines macros `True` and `False` which conflict with ids used inside vulkan.hpp


charles-lunarg (approved):

mikes-lunarg:
Sadly, my include-order hack does not work for builds with `#define VK_USE_PLATFORM_XLIB_KHR`
https://github.com/KhronosGroup/Vulkan-Hpp/issues/1608

mikes-lunarg:
Superseded by #830

PR #831, opened 2023-07-17
build: Fix generator directory references

aqnuep:


charles-lunarg (approved):
Changes are appropriate, apologies for the previous refactors which likely broke these defines.

PR #833, opened 2023-07-21
build: Update to header 1.3.258

mikes-lunarg:
- Update known-good
- Modify vulkaninfo_generator.py to wrap dynamically-sized ArrayWrapper with scope brackets
- Generate source


scripts/vulkaninfo_generator.py
```diff
@@ -567,15 +567,17 @@ def PrintStructure(struct, types_to_gen, structure_names, aliases):
...
+                out += f"        for (uint32_t i = 0; i < obj.{v.arrayLength}; i++) {{\n"
                 if v.typeID in types_to_gen:
-                    out += f"        if (obj.{v.name} != nullptr) {{\n"
-                    out += f"            p.SetElementIndex(i);\n"
-                    out += f'            Dump{v.typeID}(p, "{v.name}", obj.{v.name}[i]);\n'
-                    out += f"        }}\n"
+                    out += f"            if (obj.{v.name} != nullptr) {{\n"
+                    out += f"                p.SetElementIndex(i);\n"
+                    out += f'                Dump{v.typeID}(p, "{v.name}", obj.{v.name}[i]);\n'
+                    out += f"            }}\n"
```
charles-lunarg, line 577: automatic code formatting? Not sure if this is your work or you didn't notice it happening. (Not holding up the PR, this is a nitpick)

charles-lunarg (approved):

scripts/vulkaninfo_generator.py
mikes-lunarg: I added another set of braces around this code so indented everything inside. Here is what the output looks like: https://github.com/KhronosGroup/Vulkan-Tools/pull/833/files#diff-37d4883478e363cbe28b8d0750857d62686f218902043f53f263a51833d509dbR1972-R1980

charles-lunarg: Ahhhhh 🤦 thats a bug on my part. ArrayWrapper will 'close' the output at the end of scope, which means braces need to be minded. Which they weren't, causing bad output formatting. Great catch!

PR #836, opened 2023-07-26
pMessageIdName maybe NULL.

water-chika:


charles-lunarg (approved):

charles-lunarg:
Just commenting (since I forgot to put a message in the review) - Thanks for the fix! I checked the spec and indeed the pMessageIdName can be null. Thats a rather nasty oversight.

PR #841, opened 2023-08-20
Support tests on Windows & Use Vulkan-Loader for vulkan-1.lib

charles-lunarg:
Previously, the vulkan-1.lib file was checked into the tree, so that it wasn't
a required build step for windows builds. This means that to use functions in
new versions of vulkan, the vulkan-1.lib file would have to be updated, which
is annoying.

This commit makes the necessary changes to allow linking to the vulkan-1.lib
built by Vulkan-Loader directly, as well as adding Vulkan-Loader as a
dependency for windows builds. It was previously only built for linux & macOS.

With the use of the Vulkan-Loader, the FindVulkan.cmake file could be removed,
as well as the checked in vulkan-1.lib files, one each for x86 and x64.

This PR also enables running of MockICD tests on windows, implementing the
necessary fixes that were previously blocking it.


CMakeLists.txt
```diff
@@ -53,28 +53,15 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
...
-#  Env vars VULKAN_HEADERS_INSTALL_DIR and/or VULKAN_LOADER_INSTALL_DIR
-#  If on MacOS
-#   CMake option MOTLENVK_REPO_ROOT
-#   Env vars MOLTENVK_REPO_ROOT
-#  Fallback to FindVulkan operation using SDK install or system installed components.
-# ~~~
 set(VULKAN_HEADERS_INSTALL_DIR "HEADERS-NOTFOUND" CACHE PATH "Absolute path to a Vulkan-Headers install directory")
+if (VULKAN_HEADERS_INSTALL_DIR STREQUAL "HEADERS-NOTFOUND")
+    message(FATAL_ERROR "Unable to locate required Vulkan Headers!")
+endif()
```
juan-lunarg, line 59: System package managers don't use the INSTALL_DIR variables. They just use CMAKE_PREFIX_PATH.

```suggestion
```

CMakeLists.txt
```diff
@@ -53,28 +53,15 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
...
 set(VULKAN_LOADER_INSTALL_DIR "LOADER-NOTFOUND" CACHE PATH "Absolute path to a Vulkan-Loader install directory")
-if(WIN32 AND "${VULKAN_LOADER_INSTALL_DIR}" STREQUAL "LOADER-NOTFOUND")
-    if(CMAKE_CL_64)
-        set(VULKAN_LOADER_INSTALL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/x64")
-    else()
-        set(VULKAN_LOADER_INSTALL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/x86")
-    endif()
+if (VULKAN_LOADER_INSTALL_DIR STREQUAL "LOADER-NOTFOUND")
+    message(FATAL_ERROR "Unable to locate required Vulkan Loader!")
 endif()
```
juan-lunarg, line 64:
```suggestion
```

CMakeLists.txt
```diff
@@ -87,13 +74,27 @@ if(APPLE)
...
-message(STATUS "Using find_package to locate Vulkan")
-find_package(Vulkan)
+
 find_package(VulkanHeaders REQUIRED CONFIG)
-get_filename_component(Vulkan_LIBRARY_DIR ${Vulkan_LIBRARY} DIRECTORY)
-message(STATUS "Vulkan FOUND = ${Vulkan_FOUND}")
-message(STATUS "Vulkan Lib Dir = ${Vulkan_LIBRARY_DIR}")
-message(STATUS "Vulkan Lib = ${Vulkan_LIBRARY}")
+
+# Find and create the Vulkan::Vulkan library using the VULKAN_LOADER_INSTALL_DIR for the vulkan loader and VULKAN_HEADERS_INSTALL_DIR for the include
```
juan-lunarg, line 80: If you add VULKAN_LOADER_INSTALL_DIR to CMAKE_PREFIX_PATH the code becomes simlper. And works better for system package managers.

```
    find_library(Vulkan_LIBRARY NAMES vulkan-1 vulkan)
```

Note: Double check if the PATH_SUFFIXES part is needed for windows.

CMakeLists.txt
```diff
@@ -87,13 +74,27 @@ if(APPLE)
...
+if(WIN32)
+    find_library(Vulkan_LIBRARY
+        NAMES vulkan-1
+        HINTS
+            "${VULKAN_LOADER_INSTALL_DIR}/lib"
+            "${VULKAN_LOADER_INSTALL_DIR}/bin"
```
juan-lunarg, line 86: Which library are we looking for? The static or dynamic library?

CMakeLists.txt
```diff
@@ -87,13 +74,27 @@ if(APPLE)
...
+else()
+    find_library(Vulkan_LIBRARY
+      NAMES vulkan
+      HINTS
+        "${VULKAN_LOADER_INSTALL_DIR}/lib")
+endif()
+add_library(Vulkan::Vulkan UNKNOWN IMPORTED)
+set_target_properties(Vulkan::Vulkan PROPERTIES
+    IMPORTED_LOCATION "${Vulkan_LIBRARY}"
+    INTERFACE_INCLUDE_DIRECTORIES "${VULKAN_HEADERS_INSTALL_DIR}/include")
```
juan-lunarg, line 97: Replace with `target_link_libraries(Vulkan::Vulkan INTERFACE Vulkan::Headers)`

tests/CMakeLists.txt
```diff
@@ -35,14 +35,19 @@ find_package(GTest REQUIRED CONFIG QUIET)
...
 target_sources(vulkan_tools_tests PRIVATE
     main.cpp
+    test_common.h
     icd/mock_icd_tests.cpp
 )
 get_target_property(TEST_SOURCES vulkan_tools_tests SOURCES)
 source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}" FILES ${TEST_SOURCES})
 
-target_include_directories(vulkan_tools_tests PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+target_include_directories(vulkan_tools_tests PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
```
juan-lunarg, line 44: Do we need CMAKE_CURRENT_BINARY_DIR?

CMakeLists.txt
charles-lunarg: Ostensibly the static library, but thats only for compilation. To *run* the tests, the dynamic library is needed, and will have to be copied over to the tests folder. This is not done currently, and was an omission on my part. (everything worked on my dev machine cause I have the loader installed to system32).

tests/CMakeLists.txt
charles-lunarg: The test_common.h header is located in the current source directory, while there is a subfolder for each thing to test (currently only Mock ICD). Since the mock_icd_tests.cpp are in that subfolder, it needs to be able to find the header located in the parent directory.

If there is a better way to do this, don't hesitate to mention it.

juan-lunarg: It's fine. I have just seen CMAKE_CURRENT_BINARY_DIR copy pasted more than anything and was wondering if it could be removed.

charles-lunarg: The CMAKE_CURRENT_BINARY_DIR is so that test code can access the built executables through a generated header that contains the paths to the exes. This capability is straight up copied from the loader FYI. Any deficiencies in the loader's implementation (like not refreshing after switching compilers, bitness, build-types, etc) would be present here.

CMakeLists.txt
```diff
@@ -87,13 +70,25 @@ if(APPLE)
...
+
 find_package(VulkanHeaders REQUIRED CONFIG)
-get_filename_component(Vulkan_LIBRARY_DIR ${Vulkan_LIBRARY} DIRECTORY)
-message(STATUS "Vulkan FOUND = ${Vulkan_FOUND}")
-message(STATUS "Vulkan Lib Dir = ${Vulkan_LIBRARY_DIR}")
-message(STATUS "Vulkan Lib = ${Vulkan_LIBRARY}")
+
+# Find and create the Vulkan::Vulkan library
+if (WIN32)
+    find_program(Vulkan_LIBRARY NAMES vulkan-1.dll)
```
juan-lunarg, line 78: Since this code is only needed for testing move it into `tests/CMakeLists.txt`

CMakeLists.txt
```diff
@@ -53,28 +53,11 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
...
-#  User-supplied CMAKE_PREFIX_PATH containing paths to the header and/or loader install dirs
-#  CMake options VULKAN_HEADERS_INSTALL_DIR and/or VULKAN_LOADER_INSTALL_DIR
-#  Env vars VULKAN_HEADERS_INSTALL_DIR and/or VULKAN_LOADER_INSTALL_DIR
-#  If on MacOS
-#   CMake option MOTLENVK_REPO_ROOT
-#   Env vars MOLTENVK_REPO_ROOT
-#  Fallback to FindVulkan operation using SDK install or system installed components.
-# ~~~
 set(VULKAN_HEADERS_INSTALL_DIR "HEADERS-NOTFOUND" CACHE PATH "Absolute path to a Vulkan-Headers install directory")
 set(VULKAN_LOADER_INSTALL_DIR "LOADER-NOTFOUND" CACHE PATH "Absolute path to a Vulkan-Loader install directory")
```
juan-lunarg, line 57: This code doesn't need to exist

```suggestion
```

CMakeLists.txt
charles-lunarg: Do you mean the creating of `Vulkan::Vulkan` entirely, or just the parts that find `vulkan-1.dll` specifically? I am 99% sure that vkcube links to `vulkan-1.lib` so that part is necessary even when tests aren't being built.

juan-lunarg: just the parts that find vulkan-1.dll specifically

charles-lunarg:
Tests are failing to run because the vulkan-loader ignores VK_DRIVER_FILES as actions runs in admin mode - causing the Mock ICD driver to not be found. 

Not sure what the solution is, since the only way to get the loader to find things when running in admin mode is to modify the registry. Which is pretty ugly if you ask me, but it is what it is.

juan-lunarg:
> Not sure what the solution is, since the only way to get the loader to find things when running in admin mode is to modify the registry. Which is pretty ugly if you ask me, but it is what it is.

What about `LOADER_USE_UNSAFE_FILE_SEARCH`?

charles-lunarg:
Thats totally an option! Thanks for reminding me - I'll address your comments and repush.

CMakeLists.txt
```diff
@@ -53,28 +53,7 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
...
-if(WIN32 AND "${VULKAN_LOADER_INSTALL_DIR}" STREQUAL "LOADER-NOTFOUND")
-    if(CMAKE_CL_64)
-        set(VULKAN_LOADER_INSTALL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/x64")
-    else()
-        set(VULKAN_LOADER_INSTALL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/x86")
-    endif()
-endif()
-set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH};${VULKAN_HEADERS_INSTALL_DIR};${VULKAN_LOADER_INSTALL_DIR};
-    $ENV{VULKAN_HEADERS_INSTALL_DIR};$ENV{VULKAN_LOADER_INSTALL_DIR})
+list(APPEND CMAKE_PREFIX_PATH ${VULKAN_HEADERS_INSTALL_DIR} ${VULKAN_LOADER_INSTALL_DIR})
```
juan-lunarg, line 56: Note this can be removed entirely. See scripts/CMakeLists.txt

CMakeLists.txt
```diff
@@ -87,13 +66,15 @@ if(APPLE)
...
-message(STATUS "Using find_package to locate Vulkan")
-find_package(Vulkan)
+
 find_package(VulkanHeaders REQUIRED CONFIG)
-get_filename_component(Vulkan_LIBRARY_DIR ${Vulkan_LIBRARY} DIRECTORY)
-message(STATUS "Vulkan FOUND = ${Vulkan_FOUND}")
-message(STATUS "Vulkan Lib Dir = ${Vulkan_LIBRARY_DIR}")
-message(STATUS "Vulkan Lib = ${Vulkan_LIBRARY}")
+
+# Find and create the Vulkan::Vulkan library
```
juan-lunarg, line 70: This comment isn't very helpful.

tests/CMakeLists.txt
```diff
@@ -35,19 +50,30 @@ find_package(GTest REQUIRED CONFIG QUIET)
...
+endif()
+set_target_properties(vulkan_tools_tests PROPERTIES MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
 
 if (ENABLE_ADDRESS_SANITIZER)
     target_compile_options(vulkan_tools_tests PUBLIC -fsanitize=address)
     target_link_options(vulkan_tools_tests PUBLIC -fsanitize=address)
 endif ()
 
+if (WIN32)
+# Copy the loader shared lib (if built) to the test application directory so the test app finds it.
```
juan-lunarg, line 73: Indentation issue

CMakeLists.txt
charles-lunarg: D'oh!
Yeah, completely redundant.

juan-lunarg (commented):
Minor nitpicks

CMakeLists.txt
juan-lunarg: What I mean is Vulkan::Vulkan only makes sense as a target name if you are aware of the existing Vulkan support built into to CMake

juan-lunarg: I think something like vulkan_loader is much more clear. This one is very much an opinion though.

But this comment really doesn't say anything that the code below doesn't.

charles-lunarg: The only hiccup I have changing Vulkan::Vulkan to vulkan_loader is that it *also* includes the headers, which isn't obvious from the name vulkan_loader. Happy to change Vulkan::Vulkan to something more descriptive, it is confusing to use the same name as the existing CMake vulkan support. I can totally see myself seeing Vulkan::Vulkan in this repo and coming to the wrong conclusion.

juan-lunarg: Maybe I'm just overthinking it. Looks good to me overall.

juan-lunarg (approved):

tests/CMakeLists.txt
charles-lunarg: Fixed.

PR #842, opened 2023-08-20
vulkaninfo: Make -j=<GPU> work

### LINKED ISSUE #840: vulkaninfo cannot get info on GPU 1 only
Hi,

Unless I completely misunderstand how to use the `vulkaninfo` tool on the command line, there's something strange afoot here..
If I run the tool without any arguments, I get a list like this:
```
Layers: count = 4
=================
VK_LAYER_INTEL_nullhw (INTEL NULL HW) Vulkan version 1.1.73, layer version 1:
	Layer Extensions: count = 0
	Devices: count = 3
		GPU id = 0 (AMD Radeon Graphics (RADV REMBRANDT))
		Layer-Device Extensions: count = 0

		GPU id = 1 (NVIDIA GeForce RTX 3070 Ti Laptop GPU)
		Layer-Device Extensions: count = 0

		GPU id = 2 (llvmpipe (LLVM 15.0.7, 256 bits))
		Layer-Device Extensions: count = 0

[...]
```

According to the help message (`vulkaninfo --help`) I should be able to get info on a select GPU:
```
[-j=<gpu-number>, --json=<gpu-number>]
                    For a multi-gpu system, a single gpu can be targetted by
                    specifying the gpu-number associated with the gpu of 
                    interest. This number can be determined by running
                    vulkaninfo without any options specified.
```

And then I try to run it like this:
```
vulkaninfo -j=1
```

-- but I only get the help message, alluring the fact that I'm possibly using the tool in the wrong way. And I do not want a JSON output, but rather text, which is more easily readable.

Platform is Ubuntu 23.04.

charles-lunarg:
The help text indicates that -j=<GPU> should work, but was never implemented when the code was refactored.

Partially solves #840

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -913,9 +913,13 @@ util::vulkaninfo_optional<ParsedResults> parse_arguments(int argc, char **argv, 
 #endif
                 ++i;
             }
-        } else if (strncmp("--json", argv[i], 6) == 0 || strcmp(argv[i], "-j") == 0) {
+        } else if (strncmp("--json", argv[i], 6) == 0 || strncmp(argv[i], "-j", 2) == 0) {
             if (strlen(argv[i]) > 7 && strncmp("--json=", argv[i], 7) == 0) {
                 results.selected_gpu = static_cast<uint32_t>(strtol(argv[i] + 7, nullptr, 10));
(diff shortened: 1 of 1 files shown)


juan-lunarg (approved):

charles-lunarg:
I went ahead and cherry picked it onto the sdk-1.3.261 branch. Small fix but its good to have it in.

PR #860, opened 2023-09-19
icd: Add VkPhysicalDeviceDriverProperties

### LINKED ISSUE #788: VK_LAYER_TESTS_PRINT_DRIVER doesn't work properly with mock icd
Make it print that the mock icd name and commit

charles-lunarg:
This commit adds support for the VkPhysicalDeviceDriverProperties struct in the output of vkGetPhysicalDeviceProperties2. This is useful for users of Mock ICD wishing to know the version of MockICD used. The driverInfo field is used for this purpose and contains the git branch & tag info data. The use of Git in CMake is how the information is seamlessly gotten and provided to the source code in the form of compile definitions.

Fixes #788

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,27 @@ endif()
 # Default to using the static CRT
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
+# Find the Git branch & tag info for use in Mock ICD
+find_package (Git)
+if (GIT_FOUND AND EXISTS "${CMAKE_CURRENT_LIST_DIR}/.git/HEAD")
+    execute_process(
+        COMMAND ${GIT_EXECUTABLE} describe --tags --always
+        WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+        OUTPUT_VARIABLE GIT_TAG_INFO)
+    string(REGEX REPLACE "\n$" "" GIT_TAG_INFO "${GIT_TAG_INFO}")
+
+    file(READ "${CMAKE_CURRENT_LIST_DIR}/.git/HEAD" GIT_HEAD_REF_INFO)
+    if (GIT_HEAD_REF_INFO)
+        string(REGEX MATCH "ref: refs/heads/(.*)" _ ${GIT_HEAD_REF_INFO})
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -139,6 +139,29 @@ endif()
 # Default to using the static CRT
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
+# Find the Git branch & tag info for use in Mock ICD
+set(GIT_BRANCH_NAME "--unknown--")
+set(GIT_TAG_INFO "--unknown--")
+find_package (Git)
+if (GIT_FOUND AND EXISTS "${CMAKE_CURRENT_LIST_DIR}/.git/HEAD")
```
juan-lunarg, line 146: Your CMake and C++ code don't match intent.

The CMake code will pass even if it doesn't find git or the .git folder.

Yet the C++ code requires GIT_BRANCH_NAME and GIT_TAG_INFO be defined.

IE make the C++ code match the permissive nature of the CMake code.

charles-lunarg: The `GIT_BRANCH_NAME` and `GIT_TAG_INFO` variables in CMake are always defined, they will just default to `--unknown--`. This allows the lack of a .git folder to gracefully fallback to predefined values.

Now, you could argue that the `target_compile_definitions(VkICD_mock_icd PRIVATE GIT_BRANCH_NAME="${GIT_BRANCH_NAME}" GIT_TAG_INFO="${GIT_TAG_INFO}")` should be hidden behind an if, as well as the C++ code behind a `#if`.

juan-lunarg: I was mainly trying to articulate that this will break the gn build. Due to the C++ code relying on the CMake code.

charles-lunarg: Oh. You should have said that then. I read that as just 'an inconsistency of intent'.

charles-lunarg: I have added a #if defined() #else #endif to handle the lack of definition in the code. Thanks for clearing up my misunderstanding.

charles-lunarg:
Okay, I fixed the slew of self inflicted compiler & build & test errors. It is good to review in full now.

juan-lunarg (approved):

PR #863, opened 2023-09-22
cmake: Fix ICD json logic

### LINKED ISSUE #861: icd/CMakeLists.txt should not use copy command when targeting WIN32
In icd/CMakeLists.txt, there are statements like:
```
if (WIN32)
  add_custom_target(json COMMAND copy src_json dst_json)
endif()
```

In CMake code, the ```CMAKE_SYSTEM_XXX``` variables always describe the target platform. The same is true for the short ```WIN32```, ```UNIX```, ```APPLE``` variables. (https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html)
WIN32 do not mean host platform is Windows rather than means target platform is Windows. 
And copy command is a windows command. It should only be used on Windows platform. It should be used like:
```
if (CMAKE_HOST_WIN32)
  add_custom_target(json COMMAND copy src_json dst_json)
endif()
```

I think origin statements could be changed to:
```
if (WIN32)
  configure_file(src_json dst_json COPYONLY)
  add_custom_target(json DEPENDS dst_json)
endif()
```

juan-lunarg:
closes #861


juan-lunarg:
> CI Vulkan-Tools build # 1151 failed.

Something strange is going on with the Jenkins CI. Merging as is.

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\BuildCustomizations\masm.targets(70,5): error MSB4175: The task factory "XamlTaskFactory" could not be loaded from the assembly "Microsoft.Build.Tasks.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". Could not find file 'C:\Users\LunarG\AppData\Local\Temp\qv5ovz3w.dll'. [D:\j\w4\64-bit\Release\Vulkan-Tools\external\Vulkan-Loader\build\loader\loader-unknown-chain.vcxproj]
```

PR #867, opened 2023-09-27
gn: Fix Chromium build

### LINKED ISSUE #866: ANGLE rolls breaking on Windows after ICD json change
Hello,
After the change regarding the ICD json logic update (9616098f617c4a53ecb72a1456bb229989e2d0e4), we see that ANGLE rolls fail on some tests on Windows with the following error:
```
RendererVk.cpp:191 (VerifyExtensionsPresent): Extension not supported: VK_KHR_surface
RendererVk.cpp:191 (VerifyExtensionsPresent): Extension not supported: VK_KHR_win32_surface
Display.cpp:1052 (initialize): ANGLE Display::initialize error 0: Internal Vulkan error (-7): A requested extension is not supported, in ..\..\src\libANGLE\renderer\vulkan\RendererVk.cpp, enableInstanceExtensions:1698.
```
The output JSON file (`VkICD_mock_icd.json`) has been changed to include the following:
```
"library_path": "..\\VkICD_mock_icd.dll",
```
If the line above is changed to `.\\VkICD_mock_icd.dll`, the failing tests from above pass.

CC: @juan-lunarg

juan-lunarg:
closes #866

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -83,6 +83,10 @@ if (!is_android) {
       rebase_path(vulkan_data_dir, root_build_dir)
     ] + rebase_path(sources, root_build_dir)
 
+    if (is_win) {
+      args += [ "--no-path-prefix" ]
+    }
+
     data = outputs
   }
 }


charles-lunarg (approved):

build-gn/generate_vulkan_layers_json.py
```diff
@@ -88,7 +88,8 @@ def main():
     if args.no_path_prefix:
         relative_path_prefix = ''
     elif args.platform == 'Windows':
-        relative_path_prefix = r'..\\'  # json-escaped, hence two backslashes.
+        # https://github.com/KhronosGroup/Vulkan-Tools/issues/866
+        relative_path_prefix = r'.\\'  # json-escaped, hence two backslashes.
```
juan-lunarg, line 92: @mikes-lunarg I feel like my fix is a hack since it deviates from the script in VVL. Also it doesn't account for why linux is working which also uses relative paths.

Not sure where I went wrong with the chromium build logic since I really just tried to copy VVL.

BUILD.gn
```diff
@@ -83,6 +83,10 @@ if (!is_android) {
       rebase_path(vulkan_data_dir, root_build_dir)
     ] + rebase_path(sources, root_build_dir)
 
+    if (is_win) {
+      args += [ "--no-path-prefix" ]
+    }
+
```
juan-lunarg, line 89: Before my ICD cleanup commit the json didn't have a relative path, just "VkICD_mock_icd.dll"

So this should keep the behavior Angle was expecting.

PR #868, opened 2023-09-29
cmake: General cleanup

juan-lunarg:
- APPEND CMAKE_MODULE_PATH
- Remove overly specific CMake code
- Use CMAKE_SYSTEM_NAME to detect Linux/BSD
- Remove unused code
- Simplify install target code
- Use PRIVATE instead of PUBLIC
- Prefer target based commands


charles-lunarg (approved):
Cleaner and nicer to boot!

PR #874, opened 2023-10-05
vulkaninfo: Fix printing arrays of Enums

charles-lunarg:
Previously, the code generator assumed enums were always printed as key-value pairs, and not in an array. This led to using object syntax inside of arrays in the JSON output, which is invalid, and is now fixed.

### DIFF
--- a/scripts/vulkaninfo_generator.py
+++ b/scripts/vulkaninfo_generator.py
@@ -601,10 +601,13 @@ def PrintStructure(struct, types_to_gen, structure_names, aliases):
                 out += f'        ArrayWrapper arr(p,"{v.name}", obj.' + v.arrayLength + ');\n'
                 out += f"        for (uint32_t i = 0; i < obj.{v.arrayLength}; i++) {{\n"
                 if v.typeID in types_to_gen:
-                    out += f"            if (obj.{v.name} != nullptr) {{\n"
-                    out += f"                p.SetElementIndex(i);\n"
-                    out += f'                Dump{v.typeID}(p, "{v.name}", obj.{v.name}[i]);\n'
-                    out += f"            }}\n"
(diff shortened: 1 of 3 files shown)


richard-lunarg (approved):

mikes-lunarg:
I tried testing with the mock ICD which looks like it populates `VkPhysicalDeviceHostImageCopyPropertiesEXT`, any idea what is going on with the empty lists here?
```
"VkPhysicalDeviceHostImageCopyPropertiesEXT": {
	"copySrcLayoutCount": 5,
	"pCopySrcLayouts": [

	],
	"copyDstLayoutCount": 5,
	"pCopyDstLayouts": [

	],
	"optimalTilingLayoutUUID": [
		9,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0,
		0
	],
	"identicalMemoryTypeRequirements": false
},
```

charles-lunarg:
So... `VkPhysicalDeviceHostImageCopyPropertiesEXT` is required to be called twice. Once to get the count, another time to get the data filled out into the `pCopySrcLayouts` and `pCopyDstLayouts` members. So vulkaninfo isn't doing that part of the query, and thus the printing never reads the values (since `pCopySrcLayouts` and `pCopyDstLayouts` are null pointers).

charles-lunarg:
@mikes-lunarg I have fixed VkPhysicalDeviceHostImageCopyPropertiesEXT to use a hand written get props instead of autogen because making it possible for the pNext chain to do the call twice is a lot more work than this one special case, in addition to the difficulty in determining which structs have dynamic array members.

mikes-lunarg (approved):
Interesting, and nice workaround! For some reason the double-tap pattern seems way more awkward when it involves a pNext chain.

PR #876, opened 2023-10-05
cmake: Use XCODE variable

juan-lunarg:
Simplifies CMake code in various places for Apple builds

### DIFF
--- a/cube/macOS/cube/cube.cmake
+++ b/cube/macOS/cube/cube.cmake
@@ -28,14 +28,14 @@ set(cube_RESOURCES ${CMAKE_BINARY_DIR}/staging-json/MoltenVK_icd.json
     ${CMAKE_CURRENT_SOURCE_DIR}/macOS/cube/Resources/VulkanIcon.icns)
 
 # Have Xcode handle the Storyboard
-if(${CMAKE_GENERATOR} MATCHES "^Xcode.*")
+if(XCODE)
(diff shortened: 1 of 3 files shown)


charles-lunarg (approved):
Seems straight forward to me, but I didn't look into whether CMake uses XCODE for all apple platforms or not, so take this approval with a big grain of salt.

richard-lunarg:
Is this just an update for when you use the XCode generator like this?

cmake -G Xcode

Will the above XCode generator still work, or is this just a newer way to encode this in the cmake file?

cube/macOS/cube/cube.cmake
```diff
@@ -28,14 +28,14 @@ set(cube_RESOURCES ${CMAKE_BINARY_DIR}/staging-json/MoltenVK_icd.json
     ${CMAKE_CURRENT_SOURCE_DIR}/macOS/cube/Resources/VulkanIcon.icns)
 
 # Have Xcode handle the Storyboard
-if(${CMAKE_GENERATOR} MATCHES "^Xcode.*")
```
juan-lunarg, line 31: Instead of checking CMake generator for Xcode it's simpler to just use the established XCODE variable.

https://cmake.org/cmake/help/latest/variable/XCODE.html#xcode

juan-lunarg:
> Will the above XCode generator still work, or is this just a newer way to encode this in the cmake file?

The XCode generator will still work. This just simplifies the existing CMake code

PR #884, opened 2023-10-10
cmake: Use pkg-config for Wayland

juan-lunarg:


cube/CMakeLists.txt
```diff
@@ -33,10 +33,13 @@ if (CMAKE_SYSTEM_NAME MATCHES "Linux|BSD")
...
-        find_package(WaylandScanner REQUIRED)
-        find_package(WaylandProtocols REQUIRED)
-        include_directories(${WAYLAND_CLIENT_INCLUDE_DIR})
+        pkg_check_modules(WAYLAND_CLIENT REQUIRED IMPORTED_TARGET wayland-client)
+
+        pkg_get_variable(WAYLAND_SCANNER_EXECUTABLE wayland-scanner wayland_scanner)
+        message(STATUS "WAYLAND_SCANNER_EXECUTABLE = ${WAYLAND_SCANNER_EXECUTABLE}")
+
+        pkg_get_variable(WAYLAND_PROTOCOLS_PATH wayland-protocols pkgdatadir)
+        message(STATUS "WAYLAND_PROTOCOLS_PATH = ${WAYLAND_PROTOCOLS_PATH}")
```
juan-lunarg, line 42: Replace the need for the 3 FindWayland*.cmake modules.

charles-lunarg (approved):
Good cleanup. No need to have our own FindBlah.cmake files when they are already on the system.

PR #885, opened 2023-10-10
cmake: vkcube/vulkaninfo successfully compile for Android

juan-lunarg:


cube/CMakeLists.txt
```diff
@@ -171,40 +171,53 @@ endif()
...
-        target_link_libraries(vkcube rt)
+        target_link_libraries(vkcube PRIVATE rt)
     endif()
-
-    if (ENABLE_ADDRESS_SANITIZER)
-        target_compile_options(vkcube PUBLIC -fsanitize=address)
-        target_link_options(vkcube PUBLIC -fsanitize=address)
-    endif ()
+    target_link_libraries(vkcube PRIVATE Vulkan::Headers Vulkan::Loader)
 else()
```
charles-lunarg, line 199:
```suggestion
elseif(WIN32)
```

charles-lunarg: Might as well make it specific to WIN32 - in case some other platform wants to be added in the future.

juan-lunarg:
Not sure why but MacOS started complaining about sprintf...

```
/Users/runner/work/Vulkan-Tools/Vulkan-Tools/cube/cube.c:529:5: error: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Werror,-Wdeprecated-declarations]
    sprintf(message, "%s - Message Id Number: %d | Message Id Name: %s\n\t%s\n", prefix, pCallbackData->messageIdNumber,
```

cube/CMakeLists.txt
juan-lunarg: 👍🏾

charles-lunarg (approved):
I'm good with these changes - assuming CI passes of course. Android is a complex beast so I commend you for tackling it.

PR #902, opened 2023-10-27
build: Fix MinGW build

juan-lunarg:


CMakeLists.txt
```diff
@@ -100,6 +100,11 @@ elseif(MSVC)
     add_compile_options("/w34057")
     # Warn about signed/unsigned mismatch.
     add_compile_options("/w34245")
+
+    # Control flow guard
+    # TODO: Why do we have this? Nothing in this repo is security critical.
+    add_compile_options(/guard:cf)
```
charles-lunarg, line 106: If I had to guess it’s because vulkaninfo ships with the vulkan runtime and needs to meet certain criteria that windows drivers have to pass

juan-lunarg: Ah I didn't know that. I'll update the code to reflect that. Looks like you are correct from what I can tell.

cube/CMakeLists.txt
```diff
@@ -84,13 +84,6 @@ endif()
 
 if(WIN32)
     add_definitions(-DVK_USE_PLATFORM_WIN32_KHR -DWIN32_LEAN_AND_MEAN)
-    if(MSVC AND NOT MSVC_VERSION LESS 1900)
-        # If MSVC, Enable control flow guard
-        message(STATUS "Building vkcube with control flow guard")
-        add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/guard:cf>")
-        set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /guard:cf")
-        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /guard:cf")
-    endif()
```
juan-lunarg, line 93: cube/cubepp has no need for this.

vulkaninfo/CMakeLists.txt
```diff
@@ -114,18 +114,19 @@ if (MINGW)
     target_link_libraries(vulkaninfo PRIVATE ucrt)
 endif ()
 
+if (MSVC)
+    # vulkaninfo ships with the vulkan runtime and needs to meet certain criteria that windows drivers have to pass
```
juan-lunarg, line 118: Updated based on info from @charles-lunarg

charles-lunarg: I mean that was more of a guess rather than a hard reason.

juan-lunarg: I can just remove it entirely and add it back in if we realize we need it.

charles-lunarg: That works for me

charles-lunarg (approved):

PR #903, opened 2023-10-27
cmake: Cleanup MinGW specific code

juan-lunarg:


charles-lunarg (approved):
Seems like Ming’s doesn’t need the def file which is nice.

icd/CMakeLists.txt
```diff
@@ -75,8 +75,6 @@ if(MSVC)
     target_compile_options(VkICD_mock_icd PRIVATE /bigobj)
     target_compile_definitions(VkICD_mock_icd PRIVATE _CRT_SECURE_NO_WARNINGS)
     target_link_options(VkICD_mock_icd PRIVATE /DEF:${CMAKE_CURRENT_SOURCE_DIR}/${MOCK_ICD_NAME}.def)
-elseif(MINGW)
-    target_sources(VkICD_mock_icd PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/${MOCK_ICD_NAME}.def)
```
juan-lunarg, line 79: Remove implication of MINGW support for mock icd. We don't install this by default (See INSTALL_ICD). So we don't need to support this build config.

vulkaninfo/CMakeLists.txt
```diff
@@ -109,11 +109,6 @@ target_link_libraries(vulkaninfo PRIVATE
     ${CMAKE_DL_LIBS}
 )
 
-# TODO: Add MinGW CI and verify this is needed.
-if (MINGW)
-    target_link_libraries(vulkaninfo PRIVATE ucrt)
-endif ()
-
```
juan-lunarg, line 116: Turns out this isn't needed.

icd/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 # limitations under the License.
 # ~~~
 
-if (ANDROID)
+if (ANDROID OR MINGW)
```
juan-lunarg, line 19: There is no point to building the mock driver for MINGW. It's not an official deliverable of this repo (See INSTALL_ICD).

If someone complains we can add it back in.

PR #904, opened 2023-10-27
scripts: update_deps.py updates

juan-lunarg:


charles-lunarg (approved):
Bit of a chore moving these fixes across repos but it’s very much appreciated!

PR #910, opened 2023-10-31
ci: Add dependabot

juan-lunarg:


charles-lunarg (approved):
Dependabots, roll out!

PR #918, opened 2023-11-10
vulkaninfo: Autogenerate Format list

charles-lunarg:
The rewrite of vulkaninfo to use autogen did not include generating all possible VkFormats. This commit fixes that by generating the lists of formats and their corresponding extension/feature version requirements needed to correctly query their support.


vulkaninfo/vulkaninfo.cpp
```diff
@@ -533,8 +533,7 @@ void GpuDumpFeatures(Printer &p, AppGpu &gpu) {
     }
 }
 
-void GpuDumpTextFormatProperty(Printer &p, const AppGpu &gpu, PropFlags formats, std::vector<VkFormat> format_list,
-                               uint32_t counter) {
+void GpuDumpTextFormatProperty(Printer &p, const AppGpu &gpu, PropFlags formats, std::set<VkFormat> format_list, uint32_t counter) {
```
juan-lunarg, line 536: format_list is passed by value?

Looks like it can be a const reference

vulkaninfo/vulkaninfo.h
```diff
@@ -2026,13 +2000,17 @@ struct hash<PropFlags> {
...
+const std::vector<FormatRange> &get_format_ranges();
+
 // Used to sort the formats into buckets by their properties.
-std::unordered_map<PropFlags, std::vector<VkFormat>> FormatPropMap(AppGpu &gpu) {
-    std::unordered_map<PropFlags, std::vector<VkFormat>> map;
-    for (auto fmtRange : gpu.supported_format_ranges) {
-        for (int32_t fmt = fmtRange.first_format; fmt <= fmtRange.last_format; ++fmt) {
-            PropFlags pf = get_format_properties(gpu, static_cast<VkFormat>(fmt));
-            map[pf].push_back(static_cast<VkFormat>(fmt));
+std::unordered_map<PropFlags, std::set<VkFormat>> FormatPropMap(AppGpu &gpu) {
```
juan-lunarg, line 2006: Can this be passed by const ref?

vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -5212,4 +5212,19 @@ bool operator==(const VkSurfaceFormatKHR & a, const VkSurfaceFormatKHR b) {
...
+    {0, VK_EXT_YCBCR_2PLANE_444_FORMATS_EXTENSION_NAME, static_cast<VkFormat>(1000330000), static_cast<VkFormat>(1000330003)},
+    {3, nullptr, static_cast<VkFormat>(1000340000), static_cast<VkFormat>(1000340001)},
+    {0, VK_EXT_4444_FORMATS_EXTENSION_NAME, static_cast<VkFormat>(1000340000), static_cast<VkFormat>(1000340001)},
+    {3, nullptr, static_cast<VkFormat>(1000066000), static_cast<VkFormat>(1000066013)},
+    {0, VK_EXT_TEXTURE_COMPRESSION_ASTC_HDR_EXTENSION_NAME, static_cast<VkFormat>(1000066000), static_cast<VkFormat>(1000066013)},
+    {0, VK_IMG_FORMAT_PVRTC_EXTENSION_NAME, static_cast<VkFormat>(1000054000), static_cast<VkFormat>(1000054007)},
+    {0, VK_NV_OPTICAL_FLOW_EXTENSION_NAME, static_cast<VkFormat>(1000464000), static_cast<VkFormat>(1000464000)},
+    {0, VK_KHR_MAINTENANCE_5_EXTENSION_NAME, static_cast<VkFormat>(1000470000), static_cast<VkFormat>(1000470001)},
+};
+const std::vector<FormatRange>& get_format_ranges() { return format_ranges; }
```
juan-lunarg, line 5229: When the above is a constexpr std::array this getter function can be removed.

juan-lunarg (changes requested):

vulkaninfo/vulkaninfo.h
charles-lunarg: No cause its generated by the function. That would inhibit GRVO.

vulkaninfo/generated/vulkaninfo.hpp
charles-lunarg: Good point.

charles-lunarg: Actually, the way the code is currently structured prevents use of std::array, because the size has to be known, and because vulkaninfo.h is not generated, std::vector was required.

Except, only 1 function needed the format ranges in the vulkaninfo.h file, so if I move that, std::array can be used as well as removal of the get_format_ranges() function.

juan-lunarg (approved):

vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -5212,4 +5212,18 @@ bool operator==(const VkSurfaceFormatKHR & a, const VkSurfaceFormatKHR b) {
 std::ostream &operator<<(std::ostream &o, VkExtent3D &obj) {
     return o << "(" << obj.width << ',' << obj.height << ',' << obj.depth << ")";
 }
+auto format_ranges = std::array{
```
juan-lunarg, line 5215: FFTI: Add a comment explaining the contents of this array. The contents of this array aren't immediately apparent.

charles-lunarg: I would add a comment but doing so would make the rest of the code based odd choices undercommented in comparison. Rather I think the concept of a format range isn’t too complex as well as the struct definitions own comment’s explaining its existence.

PR #920, opened 2023-11-16
Remove *.cmake files and move MoltenVK logic into cube directory

juan-lunarg:
Only vkcube/vkcubepp make use of MoltenVK


charles-lunarg (approved):
Simple reorganization but still very welcome.

PR #933, opened 2023-12-06
[scripts] [gn] Fix Python 3.12 compatibility

gz83:
generate_vulkan_icd_json.py:31: SyntaxWarning: invalid escape sequence'\ '
  """Like regular glob but replaces \ with / in returned paths."""

### DIFF
--- a/scripts/gn/generate_vulkan_icd_json.py
+++ b/scripts/gn/generate_vulkan_icd_json.py
@@ -28,7 +28,7 @@ import platform
 import sys
 
 def glob_slash(dirname):
-    """Like regular glob but replaces \ with / in returned paths."""
+    r"""Like regular glob but replaces \ with / in returned paths."""
     return [s.replace('\\', '/') for s in glob.glob(dirname)]
 
 def main():


gz83:
@juan-lunarg 

PTAL thank you

scripts/gn/generate_vulkan_icd_json.py
```diff
@@ -28,7 +28,7 @@
 import sys
 
 def glob_slash(dirname):
-    """Like regular glob but replaces \ with / in returned paths."""
+    r"""Like regular glob but replaces \ with / in returned paths."""
```
juan-lunarg, line 31:
```suggestion
    # Like regular glob but replaces \ with / in returned paths.
```

juan-lunarg (commented):
Seems simpler to just turn this into a comment instead.

scripts/gn/generate_vulkan_icd_json.py
charles-lunarg: Yes but the current code is a doc string, which is an established pattern.
https://peps.python.org/pep-0257/

charles-lunarg (approved):

juan-lunarg:
Thank you for the fix @gz83 👍🏾

PR #938, opened 2023-12-14
vulkaninfo: API parameterization changes

aqnuep:
This change introduces the possibility to parameterize vulkaninfo builds based on API variant.

It is similar to changes applied to MockICD and other Vulkan ecosystem repositories in order to enable downstream Vulkan SC support.

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -2,6 +2,7 @@
  * Copyright (c) 2015-2021 The Khronos Group Inc.
  * Copyright (c) 2015-2021 Valve Corporation
  * Copyright (c) 2015-2021 LunarG, Inc.
+ * Copyright (c) 2023-2023 RasterGrid Kft.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +26,7 @@
  * Author: Charles Giessen <charles@lunarg.com>
  *
  */
+#pragma once
 
 #include <algorithm>
 #include <array>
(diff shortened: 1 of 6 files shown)


vulkaninfo/vulkaninfo.h
```diff
@@ -549,24 +553,38 @@ struct SurfaceExtension {
...
-        : major{VK_VERSION_MAJOR(version)}, minor{VK_VERSION_MINOR(version)}, patch{VK_VERSION_PATCH(version)} {}
-    std::string str() { return std::to_string(major) + "." + std::to_string(minor) + "." + std::to_string(patch); }
+class APIVersion {
+  public:
+    APIVersion() : api_version_(VK_API_VERSION_1_0) {}
+    APIVersion(uint32_t api_version) : api_version_(api_version) {}
+    APIVersion& operator=(uint32_t api_version) {
+        api_version_ = api_version;
+        return *this;
+    }
```
charles-lunarg, line 563: I find the inclusion of the copy operator strange, since it shouldn't be required here. 
This type can use rule of zero since it has no special copy/move semantics.

charles-lunarg (approved):
Changes do look good overall. One nitpick though. Also some nice cruft removal in the codegen script.

vulkaninfo/vulkaninfo.h
aqnuep: Fair point, I don't think this is needed for Vulkan anyway, I just had it there for symmetry, because on the Vulkan SC side it has custom code, but it makes sense to not include it in the Vulkan code. I've removed it.

spencer-lunarg (approved):

PR #951, opened 2024-01-12
Update volk to 1.3.275 & make vulkaninfo use volk for function loading

charles-lunarg:


vulkaninfo/vulkaninfo.cpp
```diff
@@ -935,21 +934,28 @@ const char *help_message_body =
     "[-o <filename>, --output <filename>]\n"
     "                    Print output to a new file whose name is specified by filename.\n"
     "                    File will be written to the current working directory.\n"
-    "[--text]            Produce a text version of " APP_SHORT_NAME " output to stdout. This is\n"
+    "[--text]            Produce a text version of " APP_SHORT_NAME
+    " output to stdout. This is\n"
```
lunarpapillo, line 938: are these (and below) `clang-format` changes?  They seem unrelated to the intended changes.

vulkaninfo/vulkaninfo.h
```diff
@@ -1196,7 +973,7 @@ static VkSurfaceKHR AppCreateGgpSurface(AppInstance &inst) {
     createInfo.streamDescriptor = 1;
 
     VkSurfaceKHR surface;
-    VkResult err = inst.dll.fp_vkCreateStreamDescriptorSurfaceGGP(inst.instance, &createInfo, NULL, &surface);
+    VkResult err = vkCreateStreamDescriptorSurfaceGGP(inst.instance, &createInfo, NULL, &surface);
```
lunarpapillo, line 976: This one seems unusual; I'm curious to know why we're supporting a `GGP` extension (is that Google?).

richard-lunarg (approved):
Ship it.

lunarpapillo (approved):
Code looks good; I have a couple of noob questions, but those don't need to stop this merge.

vulkaninfo/vulkaninfo.h
charles-lunarg: Because it was added at some point. GGP means google games platform, AKA Stadia.

vulkaninfo/vulkaninfo.cpp
charles-lunarg: yes, this one sneaked in under the radar.

charles-lunarg: Looked through for accidental formatting changes and reverted them - I suspect my local clang-format is "different" and don't want to think about formatting changes today.

PR #952, opened 2024-01-16
cube: Link to Threads directly 

charles-lunarg:
On Ubuntu 20.04, it was found that the validation layers fail to launch in vkcube & vkcubepp
due to a missing dependency on libpthread. While newer Ubuntu versions use a glibc version
where libpthread is integrated into libc, older ubuntu's do not so we need to link threads
directly in order for validation layers to be loadable.

### DIFF
--- a/cube/CMakeLists.txt
+++ b/cube/CMakeLists.txt
@@ -205,6 +205,15 @@ if(WIN32)
     add_compile_definitions(_USE_MATH_DEFINES)
 endif()
 
+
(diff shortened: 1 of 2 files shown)


charles-lunarg:
I'm going to go ahead and merge this since CI passed. We can always revert it if needed. This allows building a new SDK for definite confirmation that this fixes the issue.

charles-lunarg:
Wait what? The 2nd commit (updating runtime copyright) did NOT get merged???

Did I click the wrong button? No clue whats gone wrong here.

PR #957, opened 2024-02-09
vulkaninfo: Fix text format for VkPhysicalDeviceHostImageCopyPropertiesEXT

charles-lunarg:
Was missing a new line & underline of the title.


vulkaninfo/generated/vulkaninfo.hpp
```diff
@@ -1988,7 +1988,9 @@ void DumpVkPhysicalDeviceHostImageCopyPropertiesEXT(Printer &p, std::string name
     ObjectWrapper object{p, name};
     p.SetMinKeyWidth(35);
     p.PrintKeyValue("copySrcLayoutCount", obj.copySrcLayoutCount);
-    {
+    if (obj.copySrcLayoutCount == 0) {
+        p.PrintKeyValue("pCopySrcLayouts", "nullptr");
```
mikes-lunarg, line 1992: I have a slight preference for `"NULL"` since that is what the spec uses

mikes-lunarg (approved):
LGTM

vulkaninfo/generated/vulkaninfo.hpp
charles-lunarg: That makes more sense, updated.

PR #958, opened 2024-02-16
build: Update to header 1.3.278

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):
Seems like vulkan-hpp is broken somehow?

mikes-lunarg:
> Seems like vulkan-hpp is broken somehow?

Windows only, I suspect a `NOMINMAX` issue

charles-lunarg:
@mikes-lunarg You were right on the money, I quickly added the change to your PR so it'll pass CI

PR #960, opened 2024-03-01
build: Update to header 1.3.279

mikes-lunarg:
- Update known-good
- Generate source


charles-lunarg (approved):

mikes-lunarg:
> Seems that CI failed due to v1.3.279 not being a known vulkan-loader tag?

whoops!

PR #966, opened 2024-03-14
Lunarg ios fixes

richard-lunarg:
The latest MoltenVK known good has changed where the build artifacts are located. This change to the cmake files takes that into account.


charles-lunarg (approved):

charles-lunarg:
linux seems to be sporadically segfaulting, I suggest just re-running failing CI (after the mac job completes)

PR #967, opened 2024-03-15
vulkaninfo: Add quotes around NULL

charles-lunarg:
Used PrintKeyValue instead of PrintKeyString by mistake.

### DIFF
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -84,7 +84,7 @@ jobs:
                 -D BUILD_WERROR=ON \
                 -D INSTALL_ICD=ON \
                 -D BUILD_TESTS=ON \
-                -D ENABLE_ADDRESS_SANITIZER=ON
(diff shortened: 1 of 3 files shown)


jeremy-lunarg (approved):
Tested locally. Works. Thanks @charles-lunarg!

lunarpapillo (commented):
Why was this not found before?  Generator issues are typically found early.

Why did the repository tests not catch the issue?  Has a new test case been added that can detect the failure?

richard-lunarg:
@lunarpapillo These kind of failures show up from time to time, they are always specific to the hardware in use by the manual tester.

charles-lunarg:
> Why was this not found before? Generator issues are typically found early.
> 
> Why did the repository tests not catch the issue? Has a new test case been added that can detect the failure?

This specific issue is really due to PrintKeyValue not being 'protected' against printing strings, which is easy to do by mistake. 

So yes, it is annoying that it was only found due to testing on a specific platform, but is something that could have been prevented (at the compile stage, let alone the test stage)

PR #969, opened 2024-03-17
vulkaninfo: Add minImageCount/maxImageCount to VK_EXT_surface_maintenance1

HildarTheDorf:
This information from VkSurfaceCapabilitiesKHR is explicitly called out in the VK_EXT_surface_maintenance1 extension specification as something which will change based on the present mode used. It would be helpful for vulkaninfo to report it as such.


charles-lunarg (approved):
Yep! Very useful information to have, thank you for catching the omission.

PR #972, opened 2024-03-19
icd: Fix DEVICE_LOST injection null check

spencer-lunarg:
quick fix for https://github.com/KhronosGroup/Vulkan-Tools/pull/971

I swear `submitCount` had to be non-zero, but forgot it is valid to have an empty submit


artem-lunarg (approved):
LGTM

PR #977, opened 2024-03-25
cube: Poll wayland socket each iteration

### LINKED ISSUE #976: vkcube-wayland does not read from the wayland socket if un-paused
Here is the outer loop of vkcube-wayland:

```c
static void demo_run(struct demo *demo) {
    while (!demo->quit) {
        if (demo->pause) {
            wl_display_dispatch(demo->display);  // block and wait for input
        } else {
            wl_display_dispatch_pending(demo->display);  // don't block
            demo_draw(demo);
            demo->curFrame++;
            if (demo->frameCount != INT32_MAX && demo->curFrame == demo->frameCount) demo->quit = true;
        }
    }
}
```

As long as the demo is not paused, vkcube-wayland only invokes `wl_display_dispatch_pending` which will not read any data from the wayland socket. This has worked so far because the WSI would read from the wayland socket in `vkAcquireNextImageKHR`.

However, in the next mesa release, and if the wayland compositor supports it, mesa will no longer read from the wayland socket in `vkAcquireNextImageKHR`. This will cause no messages (keyboard input, resize events, etc.) to be read and dispatched.

I've used the following patch for testing, but you might come up with a more elegant solution:

```diff
diff --git a/cube/cube.c b/cube/cube.c
index e319ee8b..84ea6906 100644
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -2815,12 +2815,18 @@ static void demo_create_xcb_window(struct demo *demo) {
 }
 // VK_USE_PLATFORM_XCB_KHR
 #elif defined(VK_USE_PLATFORM_WAYLAND_KHR)
+#include <poll.h>
 static void demo_run(struct demo *demo) {
     while (!demo->quit) {
         if (demo->pause) {
             wl_display_dispatch(demo->display);  // block and wait for input
         } else {
-            wl_display_dispatch_pending(demo->display);  // don't block
+            struct pollfd pollfd = { .fd = wl_display_get_fd(demo->display), .events = POLLIN };
+            if (poll(&pollfd, 1, 0) > 0) {
+                wl_display_dispatch(demo->display);
+            } else {
+                wl_display_dispatch_pending(demo->display);  // don't block
+            }
             demo_draw(demo);
             demo->curFrame++;
             if (demo->frameCount != INT32_MAX && demo->curFrame == demo->frameCount) demo->quit = true;
```

mahkoh:
Closes #976

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -37,6 +37,7 @@
 #if defined(VK_USE_PLATFORM_XLIB_KHR) || defined(VK_USE_PLATFORM_XCB_KHR)
 #include <X11/Xutil.h>
 #elif defined(VK_USE_PLATFORM_WAYLAND_KHR)
+#include <poll.h>
 #include <linux/input.h>
 #include "xdg-shell-client-header.h"
 #include "xdg-decoration-client-header.h"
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):

charles-lunarg:
Thanks for turning this into a PR! I probably would have found this bug eventually but wouldn't have known *why* it broke nor how to fix it, as its due to upstream mesa changes. So I *really* appreciate that future pain being eviscerated.

PR #978, opened 2024-03-27
vkcube: Thread safe event polling

HildarTheDorf:
The fix for #976 has a couple of flaws, I think.
1) The poll is unnecessary as we have an always-running main loop, it's okay to do an empty read, it won't block.
2) If the driver uses multiple threads internally, we are not guarding against a concurrent read by the driver.
3) We never flush messages to the server. Again, the driver could send messages and expect the application to flush.

Both 2) and 3) assume a sufficiently naive driver, most are more defensive than the bare minimum.


charles-lunarg (commented):
I'm no wayland expert, so I'll try not to claim I understand it fully. But the changes here make sense.

If you are removing the `poll()` you should go ahead and remove the `#include <poll.h>` that was added at the top.

charles-lunarg (approved):

PR #996, opened 2024-06-04
Add support for GNU/Hurd

pinotree:
- drop an useless `#include` (simplifies the OS porting)
- cube: use the Unix `main()` also for Hurd
- build the Linux/BSD tools also on Hurd

In case of doubts: yes, `__GNU__` the official define for GNU/Hurd, see https://github.com/cpredef/predef/blob/master/OperatingSystems.md#gnu-aka-gnuhurd.


charles-lunarg (approved):
Yep, assuming this passes the CI checks, this can be merged.

PR #1000, opened 2024-06-28
build: Update to header 1.3.289

mikes-lunarg:


charles-lunarg (commented):
The autogen implementation for LayeredApiProperties is wrong - it needs a two call but the codegen doesn't handle that.

charles-lunarg:
Actually, I recommend removing the layered api properties from vulkaninfo for the time being and creating an issue to track its addition. No sense delaying header update over it.

charles-lunarg (approved):

PR #1018, opened 2024-08-05
vulkaninfo: Fix deprecation of MVK/IOS surface extensions

charles-lunarg:
These surface extensions shouldn't be enabled if the metal surface extension is present. This fixes a deprecation notice the validation layers emit - as well as not provide redundant information (since the metal extension is a direct replacement for the mvk/ios extensions).


richard-lunarg:
Strongly agree. Out with the old, in with the new.

richard-lunarg (approved):
Agree, it's time to move past these old extensions.

PR #1019, opened 2024-08-13
cube: roundtrip to ensure all globals are processed

vyivel:
wl_display_dispatch() may not process all events.


charles-lunarg (approved):
Looking at the docs, this change matches what I'm seeing - that wl_display_dispatch doesn't guarantee processing of all events. Thank you for fixing this!

charles-lunarg:
Apologies for the delay in merging, I wanted to let CI finish and forgot to hit merge...

PR #1022, opened 2024-08-23
build: Update to header 1.3.294

mikes-lunarg:


charles-lunarg (approved):
Changes appear fine even though they are causing build failures. I assume something just needs to be tweaked.

mikes-lunarg:
> Changes appear fine even though they are causing build failures. I assume something just needs to be tweaked.

Yes I needed to copy the mock vkCreatePipelineBinariesKHR from https://gitlab.khronos.org/vulkan/Vulkan-ValidationLayers/-/merge_requests/96

PR #1029, opened 2024-09-05
cube: Support runtime selection of WSI platform

charles-lunarg:
By changing the selection of a WSI platform from a build time choice to a runtime choice, this allows the removal of vkcube-wayland as a separate binary.

Use `--wsi <xlib|xcb|wayland|display|directfb>` to choose the WSI platform that vkcube will render with. The chosen platform must be one that vkcube was compiled with support for, so directfb will be unavailable without changing the build parameters to enable it.

These changes have been made to both vkcube and vkcubepp.

Platforms which do not have multiple WSI platforms, such as windows, android, and apple (metal is the only non-deprecated platform) do not have this option as there aren't any platforms to choose from.

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -26,6 +26,7 @@
  */
 
 #define _GNU_SOURCE
+#include <ctype.h>
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
@@ -33,13 +34,18 @@
 #include <stdbool.h>
 #include <assert.h>
 #include <signal.h>
+#if defined(VK_USE_PLATFORM_SCREEN_QNX)
 #include <errno.h>
-#if defined(VK_USE_PLATFORM_XLIB_KHR) || defined(VK_USE_PLATFORM_XCB_KHR)
-#include <X11/Xutil.h>
-#elif defined(VK_USE_PLATFORM_WAYLAND_KHR)
+#endif
+#if defined(VK_USE_PLATFORM_XLIB_KHR)
+#include "xlib_loader.h"
+#endif
+#if defined(VK_USE_PLATFORM_XCB_KHR)
+#include "xcb_loader.h"
+#endif
+#if defined(VK_USE_PLATFORM_WAYLAND_KHR)
 #include <linux/input.h>
-#include "xdg-shell-client-header.h"
-#include "xdg-decoration-client-header.h"
+#include "wayland_loader.h"
 #endif
 
 #ifdef _WIN32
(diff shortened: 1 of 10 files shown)


cube/cube.cpp
```diff
@@ -3394,35 +3733,84 @@ int main(int argc, char **argv) {
 
     demo.init(argc, argv);
 
+    switch (demo.wsi_platform) {
+        default:
```
enunes, line 3737: This switch needs a case for `VK_USE_PLATFORM_DISPLAY_KHR`, otherwise when `--wsi DISPLAY` is set, it falls back to the default and exits. I think nothing special needs to be done so it can just `break`. Also on cpp.

cube/cube.cpp
```diff
@@ -3394,35 +3733,84 @@ int main(int argc, char **argv) {
...
+        case (WsiPlatform::auto_):
+            fprintf(stderr,
+                    "WSI platform should have already been set, indicating a bug. Please set a WSI platform manually with "
+                    "--wsi");
```
enunes, line 3741: This error and the other similar ones can use a line break at the end so it doesn't mix with the next prompt (also on cpp).

cube/cube.c
```diff
@@ -4281,10 +4639,85 @@ static void demo_init(struct demo *demo, int argc, char **argv) {
...
         ERR_EXIT("Usage: vkcube [--validate]\n", "Usage");
 #else
+
+        // Making the help for --wsi nice requires a little extra work since the list depends on what is available at
+        // compile time
+        size_t max_str_len = 100;
+        char *available_wsi_platforms = (char *)malloc(max_str_len);
+        memset(available_wsi_platforms, 0, max_str_len);
+#if defined(VK_USE_PLATFORM_XCB_KHR)
+        strncat(available_wsi_platforms, "XCB", max_str_len);
```
enunes, line 4670: It is a bit inconsistent that in `vkcube` the options are lowercase in the help and in `vkcubepp` the options are uppercase. It does seem to work in both ways in either though. But can the help be made consistent to avoid confusion?

```
$ ./vkcube --help
	[--wsi <XCB|XLIB|WAYLAND|DISPLAY>]
$ ./vkcubepp --help
	[--wsi <xcb|xlib|wayland|display>]
```

enunes (commented):
I tested on x86_64 and aarch64 xlib, xcb, wayland and display, and this seems like a great improvement to me. Added a few comments for review. Thanks

cube/cube.c
charles-lunarg: Yes, that was a bit of a "I had to rewrite the code to use C++ and didn't notice that they were different". I went with lowercase since its simpler to type out. I kept the "tolower" transformation in case someone types it in all caps, since it doesn't hurt to do that for a user.

enunes (approved):
Looks good to me, thanks!

jjulianoatnv:
This:

>Platforms which do not have multiple WSI platforms, such as windows

is an incorrect assumption. Windows does have direct to display.

charles-lunarg:
> Platforms which do not have multiple WSI platforms, such as windows

Was such a big assumption that I found that out during development of the PR. Unfortunately, the commit message wasn't corrected after I changed the code to reflect the reality of DISPLAY being available on windows. Instead I will double check that the option is present on windows, and if not, rectify that with a subsequent PR.

charles-lunarg:
I just ran vkcube locally with the changes incorporated, both win32 and display were available as runtime options. The display option didn't work because I was running inside a VM with lavapipe as the sole driver, so if changes are needed to make display work on windows, let me know.

Otherwise I don't think there is any reasonable way to fix the git commit message without force pushing to main. That occurred because the initial version of the PR did preclude windows, but when it was addressed the commit message wasn't altered.

PR #1033, opened 2024-09-27
vulkaninfo: Correctly query per-surface queue support

charles-lunarg:
Plus other small improvements.

### DIFF
--- a/vulkaninfo/vulkaninfo.cpp
+++ b/vulkaninfo/vulkaninfo.cpp
@@ -396,17 +396,19 @@ void GpuDumpQueueProps(Printer &p, AppGpu &gpu, const AppQueueFamilyProperties &
     p.PrintKeyString("queueFlags", VkQueueFlagsString(props.queueFlags));
     p.PrintKeyValue("timestampValidBits", props.timestampValidBits);
 
-    if (queue.is_present_platform_agnostic) {
-        p.PrintKeyString("present support", queue.platforms_support_present ? "true" : "false");
+    if (!queue.can_present) {
+        p.PrintKeyString("present support", "false");
+    } else if (queue.can_always_present) {
+        p.PrintKeyString("present support", "true");
     } else {
         size_t width = 0;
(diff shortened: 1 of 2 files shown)


vulkaninfo/vulkaninfo.cpp
```diff
@@ -396,17 +396,19 @@ void GpuDumpQueueProps(Printer &p, AppGpu &gpu, const AppQueueFamilyProperties &
...
-        p.PrintKeyString("present support", queue.platforms_support_present ? "true" : "false");
+    if (!queue.can_present) {
+        p.PrintKeyString("present support", "false");
+    } else if (queue.can_always_present) {
+        p.PrintKeyString("present support", "true");
     } else {
         size_t width = 0;
-        for (auto &surface : gpu.inst.surface_extensions) {
-            if (surface.name.size() > width) width = surface.name.size();
+        for (auto &support : queue.present_support) {
```
ziga-lunarg, line 405: Could support be const for clarity?

vulkaninfo/vulkaninfo.h
```diff
@@ -243,7 +243,7 @@ auto GetVectorInit(const char *func_name, F &&f, T init, Ts &&...ts) -> std::vec
         err = f(ts..., &count, results.data());
         results.resize(count);
         iteration_count++;
-    } while (err == VK_INCOMPLETE || iteration_count < max_iterations);
+    } while (err == VK_INCOMPLETE || iteration_count >= max_iterations);
```
ziga-lunarg, line 246: What is the reason for this change? It looks wrong, could lead to infinite loop

charles-lunarg: The reason for the change was that the loop was always run 3 times. I want the loop to only run once if necessary, but I can see how it would lead to infinite loops now.

I'll have to rethink the logic some more.

ziga-lunarg (approved):

PR #1037, opened 2024-10-07
build: Update to header 1.3.297

mikes-lunarg:


charles-lunarg (approved):

charles-lunarg:
Seems that the build failed due to the Vulkan-Loader tag v1.3.297 not being available. I've restarted now that the loader PR is merged.

Issue #950, opened 2024-01-12
vkcube doesn't check that device is usable


CendioOssman:
`vkcube`'s device selection is unfortunately a bit optimistic and it blindly assumes that all devices it can enumerate are usable from its context. This is not always the case, and hence it can end up failing to start properly.

This issue showed up when running vkcube in a VNC session (TigerVNC/tigervnc#1674). In that case, the hardware device cannot be use because DRI3 isn't available. But vkcube fails to fall back to the CPU device.

I am a complete newbie when it comes to Vulkan, so I have no idea how applications are supposed to pick devices.

The noticable difference between the VNC environment and a "regular" one (according to `vulkaninfo`), is that you only get presentable surfaces for working devices. So perhaps that is a good method?

What is the de facto method for picking devices? Is there something TigerVNC can do to help applications make the right choice?

charles-lunarg:
(Apologies for a later response than I'd like, an SDK release took all available cycles)

The thing about vkcube is that it is a 1.0 application with almost no usage of exotic or poorly supported features.
This means it doesn't need to do a lot of feature checking. 

Typically, applications validates each VkPhysicalDevice for suitability before picking the device (where they use both hard requirements and heuristics) but its not uncommon to just pick the first VkPhysicalDevice.

Looking at the linked thread, I see discussion of vkcube vs vulkaninfo. Vulkaninfo doesn't pick a device, it creates a VkDevice on all available VkPhysicalDevices. It also doesn't try to create a swapchain which means it doesn't have the same requirements as vkcube.

What is the specific thing that is causing vkcube to fail? And is it something that can be checked before use?

CendioOssman:
> Looking at the linked thread, I see discussion of vkcube vs vulkaninfo. Vulkaninfo doesn't pick a device, it creates a VkDevice on all available VkPhysicalDevices. It also doesn't try to create a swapchain which means it doesn't have the same requirements as vkcube.

vulkaninfo failed because it assumed the first X11 visual it got was suitable for a top level window, which I don't think is a safe assumption. It's an issue we've seen with other applications as well, though, so we added a workaround to TigerVNC for that. vulkaninfo now works fine.

> What is the specific thing that is causing vkcube to fail? And is it something that can be checked before use?

It picks the first vulkan device, which isn't compatible with the current X11 display, and fails:

```console
$ vkcube
Selected GPU 0: Intel(R) UHD Graphics 630 (CFL GT2), type: IntegratedGpu
[...]

PR #1047, opened 2024-11-05
vkcube and vkcubepp: Check if GPU support the surface

water-chika:
Add condition to filter if the GPU support the surface.
Solve https://github.com/KhronosGroup/Vulkan-Tools/issues/950

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -4143,7 +4143,10 @@ static void demo_init_vk(struct demo *demo) {
     }
 
     volkLoadInstance(demo->inst);
+}
 
+static void demo_select_physical_device(struct demo* demo) {
+    VkResult err;
     /* Make initial call to query gpu_count, then second call for gpu info */
     uint32_t gpu_count = 0;
     err = vkEnumeratePhysicalDevices(demo->inst, &gpu_count, NULL);
@@ -4188,7 +4191,10 @@ static void demo_init_vk(struct demo *demo) {
             for (uint32_t i = 0; i < gpu_count; i++) {
                 vkGetPhysicalDeviceProperties(physical_devices[i], &physicalDeviceProperties);
                 assert(physicalDeviceProperties.deviceType <= VK_PHYSICAL_DEVICE_TYPE_CPU);
-                count_device_type[physicalDeviceProperties.deviceType]++;
+                VkBool32 supported = 0;
+                vkGetPhysicalDeviceSurfaceSupportKHR(physical_devices[i], 0, demo->surface, &supported);
+                if (supported)
+                    count_device_type[physicalDeviceProperties.deviceType]++;
             }
 
             VkPhysicalDeviceType search_for_device_type = VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU;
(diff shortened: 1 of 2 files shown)


cube/cube.c
```diff
@@ -4143,7 +4143,10 @@ static void demo_init_vk(struct demo *demo) {
     }
 
     volkLoadInstance(demo->inst);
+}
 
+static void demo_select_physical_device(struct demo* demo) {
```
water-chika, line 4148: Seperate physical device to a function because it needs create surface between creating instance and selecting physical device.

cube/cube.c
```diff
@@ -4188,7 +4191,10 @@ static void demo_init_vk(struct demo *demo) {
             for (uint32_t i = 0; i < gpu_count; i++) {
                 vkGetPhysicalDeviceProperties(physical_devices[i], &physicalDeviceProperties);
                 assert(physicalDeviceProperties.deviceType <= VK_PHYSICAL_DEVICE_TYPE_CPU);
-                count_device_type[physicalDeviceProperties.deviceType]++;
+                VkBool32 supported = 0;
```
water-chika, line 4194: Check if GPU support the surface

cube/cube.c
```diff
@@ -4492,8 +4516,6 @@ static VkSurfaceFormatKHR pick_surface_format(const VkSurfaceFormatKHR *surfaceF
 static void demo_init_vk_swapchain(struct demo *demo) {
     VkResult U_ASSERT_ONLY err;
 
-    demo_create_surface(demo);
```
water-chika, line 4495: Move demo_create_surface to time between demo_init and demo_select_physical_device.

cube/cube.c
```diff
@@ -4868,6 +4890,8 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR pCmdLine,
     demo.connection = hInstance;
     strncpy(demo.name, "Vulkan Cube", APP_NAME_STR_LEN);
     demo_create_window(&demo);
+    demo_create_surface(&demo);
```
water-chika, line 4893: Let demo_create_surface after demo_create_window

cube/cube.c
```diff
@@ -4959,6 +4985,8 @@ static void processCommand(struct android_app *app, int32_t cmd) {
                 for (int i = 0; i < argc; i++) free(argv[i]);
 
                 demo.window = (void *)app->window;
+                demo_create_surface(&demo);
```
water-chika, line 4988: Let demo_create_surface after demo.window is inited.

cube/cube.c
```diff
@@ -5044,6 +5072,8 @@ int main(int argc, char **argv) {
             break;
 #endif
     }
+    demo_create_surface(&demo);
```
water-chika, line 5075: Let demo_create_surface after window is created

charles-lunarg (commented):
Good changes.
I was wanting these changes to be made for quite a while but never found time for it. 

Only thing I would suggest doing is porting the changes over to cube.cpp so the C++ version has the same functionality.

cube/cube.c
```diff
@@ -4181,34 +4184,42 @@ static void demo_init_vk(struct demo *demo) {
...
-                search_for_device_type = VK_PHYSICAL_DEVICE_TYPE_OTHER;
-            }
+                // Continue next gpu if this gpu does not support the surface.
+                VkBool32 supported = 0;
+                vkGetPhysicalDeviceSurfaceSupportKHR(physical_devices[i], 0, demo->surface, &supported);
+                if (!supported) continue;
 
-            for (uint32_t i = 0; i < gpu_count; i++) {
-                vkGetPhysicalDeviceProperties(physical_devices[i], &physicalDeviceProperties);
-                if (physicalDeviceProperties.deviceType == search_for_device_type) {
```
water-chika, line 4209: Use priority to select physical device so that only one loop is needed. It will be more convenient to filter GPU if only one loop existed.

water-chika:
> Only thing I would suggest doing is porting the changes over to cube.cpp so the C++ version has the same functionality.

I have ported the changes over to vkcubepp.

charles-lunarg (approved):
Apologies for the delay, the changes look good.

I went ahead and added the one nitpick I did find (not checking the return value of vkGetPhysicalDeviceSurfaceSupportKHR().

PR #1053, opened 2024-11-27
cube: Link to Vulkan-Loader on macOS

charles-lunarg:
This was accidentally undone during the changes to cube to handle runtime WSI selection. By not linking, vkcube does not work unless the SDK is installed globally (ie, moltenvk is available globally).


richard-lunarg (approved):
I've verified this works! Great job and thanks!

PR #1055, opened 2024-12-02
vulkaninfo: Add video profiles support

aqnuep:
This PR accompanies https://github.com/KhronosGroup/Vulkan-Profiles/pull/726 by adding video support to the outputs of `vulkaninfo`, including profiles JSON output.

This is marked draft for now, because I'd guess we do not want to merge this until the Vulkan-Profiles change is in (and thus the profile JSON schema is updated), but it is otherwise standalone.


lunarpapillo:
The **LunarG CI** failure is due to an exposure on `main` due to an erroneous merge.  I'll re-run it once the exposure is repaired.

aqnuep:
This is now ready to be reviewed and merged as the Vulkan-Profiles changes are in.

charles-lunarg (approved):
This is a big change, but looks good. 

My comments are exactly that, comments, not changes or requests. Code looks to be just about how I would have written it.

PR #1061, opened 2024-12-16
vkcubepp: Refactor create_window and run functions

water-chika:
Use template parameter for WSI Platform to override create_window and run functions.


charles-lunarg (approved):
I love the changes - apologies for the delay in reviewing them and getting them merged. I went ahead and rebased on main which added some fuchsia code that needed to be added to your abstractions.

PR #1071, opened 2025-01-22
cube: Correctly apply sRGB OETF/EOTF

HildarTheDorf:
vkcube is incorrectly treating srgb texture values as linear, then writing linear values to a COLORSPACE_SRGB_NONLINEAR swapchain image without an _SRGB format or manual correction.

VK_FORMAT_R8G8B8A8_SRGB as a sampled image format is required to be supported, so changing the input texture is fine.

SRGB formats for swapchain are required if the equivalent UNORM is supported (and vice-versa). However it is possible for unextended vulkan 1.4 to only support 1010102 or 565 etc. formats.

This results in a slight overall change to the lighting of the cube. But it is more correct for those using vkcube's code as an example.


cube/cube.frag.inc
```diff
@@ -1,41 +1,84 @@
-	// 7.9.2888
-	0x07230203,0x00010000,0x00080007,0x00000030,0x00000000,0x00020011,0x00000001,0x0006000b,
+	// 1115.1.0
```
HildarTheDorf, line 1: Wow that's a lot of versions of glslang

charles-lunarg: I am unsure when the checked in code was compiled, because Vulkan-Tools used to compile glslang, then it went to just downloading the binary, to the current checked in state.

charles-lunarg (approved):

PR #1079, opened 2025-02-16
vkcubepp: Support Fuchsia

gnoliyil:
This change was originally in Fuchsia's Vulkan-Tools fork (https://fxrev.dev/441056). It adds Fuchsia OS
(https://fuchsia.dev) support to vkcubepp.

Fuchsia has two WSI platforms available: It can render directly to display frame buffer (`fuchsia_display`), or use Fuchsia's Scenic compositor (`fuchsia_scenic`).

All Fuchsia-specific code is under VK_USE_PLATFORM_FUCHSIA ifdef build guards and is only compiled for Fuchsia targets.

This change also adds a BUILD.gn file used for Fuchsia in-tree builds.

Test: vkcube-on-fb and vkcube-on-scenic on Fuchsia
Bug: https://fxbug.dev/378964821

### DIFF
--- a/cube/BUILD.gn
+++ b/cube/BUILD.gn
@@ -1,0 +1,152 @@
+# Copyright (C) 2025 The Fuchsia Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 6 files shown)


cube/BUILD.gn
```diff
@@ -0,0 +1,152 @@
...
+    "//sdk/lib/component/outgoing/cpp",
+    "//sdk/lib/syslog/cpp",
+    "//sdk/lib/ui/scenic/cpp",
+    "//third_party/Vulkan-Headers/src:vulkan_headers",
+    "//third_party/Vulkan-Loader:libvulkan",
+    "//zircon/system/ulib/syslog",
+  ]
+}
+
+compiled_action("glslang_frag") {
```
charles-lunarg, line 122: FYI the spirv code is checked into the repo, so the GN build doesnt need to define it. Put another way, the shader compilation step is for developers of vkcube rather than those that wish to only build it. I am going to go ahead and merge it unmodified since it doesn't harm anyone to have the glslang compilation in GN.

charles-lunarg (approved):
Lots of code, but that is to be expected when supporting a new platform.

PR #1081, opened 2025-02-16
vulkaninfo: Support Fuchsia

gnoliyil:
This change adds Fuchsia-specific build files and metadata for vulkaninfo so that it builds on Fuchsia.

Test: vulkaninfo on Fuchsia
Bug: https://fxbug.dev/378964821


charles-lunarg (approved):
Looks fine. As I don't have a fuschia system setup I will assume that the build works as expected.

PR #1083, opened 2025-02-20
vulkaninfo: Enable device groups extension

charles-lunarg:
This was accidentally disabled, which causes the device group output to dissapear.

### DIFF
--- a/vulkaninfo/vulkaninfo.h
+++ b/vulkaninfo/vulkaninfo.h
@@ -754,6 +754,9 @@ struct AppInstance {
             if (strcmp(VK_EXT_SURFACE_MAINTENANCE_1_EXTENSION_NAME, ext.extensionName) == 0) {
                 inst_extensions.push_back(ext.extensionName);
             }
+            if (strcmp(VK_KHR_DEVICE_GROUP_CREATION_EXTENSION_NAME, ext.extensionName) == 0) {
+                inst_extensions.push_back(ext.extensionName);
+            }
         }
     }
 


ziga-lunarg (approved):

spencer-lunarg (approved):

Pegasus691-69 (approved):

charles-lunarg:
I found the source of the CI failure!!!

The problem is that the CI machines use mainline ubuntu packages for the loader, and those use v1.3.275, which includes a loader commit that removes `-fno-strict-aliasing`. Turns out that is buggy, and has been fixed with:
https://github.com/KhronosGroup/Vulkan-Loader/pull/1438

But because CI uses an buggy Vulkan-Loader, the device groups code crashes.

Once CI is updated to use a newer loader version, this can safely be merged in.

lunarpapillo:
Although CI attempts to keep a variety of both new and old loaders around (to potentially catch loader bugs just like these!), a surprising number of machines use loader 1.3.275 - probably because 1.3.275 was and still remains the installed system loader on stock **Ubuntu 24.04** machines:

```
tcubumes    1.3.204
tcubumes4   1.3.204
tcubunvi    1.3.204
tcubunvi2   1.3.204
tcubuand4   1.3.275
tcubuand5   1.3.275
tcubumes7   1.3.275
tcubumes8   1.3.275
tcubumes9   1.3.275
tcubunvi3   1.3.275
tcubunvi4   1.3.275
tcubunvi5   1.3.275
tcubunvi6   1.3.275
tcubunvi7   1.3.275
tcubumes5   1.4.341
tcubumes6   1.4.341
```
The loader on the Android machines really doesn't matter, but there are still 8 machines using 1.3.275 that will have to be updated.

lunarpapillo:
All passed after rebase:

image

PR #1089, opened 2025-03-06
vulkaninfo: Use Object for "Vulkan Profiles" array

charles-lunarg:
The "Video Profiles" element must be a JSON Object in order for the sub elements to have names. The outputprinter helper doesn't prevent mixing objects inside of arrays, so this wasn't caught during development.


aqnuep:
Thanks! I'm not sure how I missed this one.

PR #1091, opened 2025-03-07
build: Update to header 1.4.310

mikes-lunarg:


charles-lunarg (approved):
Kinda surprised there isn't any changes, but I suppose it only affects NV extensions which vulkaninfo doesn't generate.

PR #1098, opened 2025-03-20
vulkaninfo: Fix unreported extension formats

andy-poole-samsung:
Changes the format support check to compare against the device extensions instead of the instance extensions.


charles-lunarg (approved):
Holy COW. Thank you for finding and fixing! This bug has been here for literal years, possibly a decade. Its so old is predates the big vulkaninfo rewrite in 2019!

charles-lunarg:
At first I thought it couldn't possibly be a real bug - often times I've found that the code was right, then became wrong on rewrite/refactor. Well, this one was very much wrong from way back when. Appreciate your colleague having some eagle eyes to spot this.

PR #1099, opened 2025-03-21
build: Update to header 1.4.311

mikes-lunarg:


spencer-lunarg (approved):
I like to believe I'm faster than @charles-lunarg at reviewing

PR #1102, opened 2025-03-27
cmake: Check wayland pkgconfig deps for cube

charles-lunarg:


charles-lunarg:
CI failed because the branch doesn't contain the fixes to vkcubepp, and all other testing succeeded.

PR #1107, opened 2025-04-07
cube: Re-record command buffers each frame

charles-lunarg:
Modifies cube & cubepp to re-record the command buffers each frame instead of pre-recording them at application startup and during resize.

This enables vast refactoring to use best practices for organizing frame and swapchain data into two resource groups: per-submission, and per- swapchain-image. A submission resource is any data needed to render (uniform buffer, descriptor set), any command buffers, the image acquire semaphore, and the fence used in the submission. There are currently two submission resource sets, and the application alternates between the two on subsequent frames. The swapchain-image resources are separate because they are dependent on the number of swapchain images and because the swapchain image index used each frame isn't deterministic.

In addition to the above architectural change, this commit refactors swapchain recreation to only recreate a minimum of necessary objects. This is the swapchain, the swapchain image views, the swapchain specific semaphores, the depth buffer, and the framebuffer. It is possible to avoid recreating frame buffers & semaphores using newer features ( dynamic_rendering & swapchain_maintenance1) but was decided against due to vkcube needing 1.0 compatability.

### DIFF
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -75,6 +75,8 @@
 
 // Allow a maximum of two outstanding presentation operations.
 #define FRAME_LAG 2
+// Need to know how big of a buffer of swapchain images, image views, and semaphores are needed
+#define MAX_SWAPCHAIN_IMAGE_COUNT 8
 
 #define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
 
@@ -388,15 +390,22 @@ const char *wsi_to_string(WSI_PLATFORM wsi_platform) {
 };
 
 typedef struct {
-    VkImage image;
+    VkFence fence;
+    VkSemaphore image_acquired_semaphore;
     VkCommandBuffer cmd;
     VkCommandBuffer graphics_to_present_cmd;
-    VkImageView view;
     VkBuffer uniform_buffer;
     VkDeviceMemory uniform_memory;
     void *uniform_memory_ptr;
-    VkFramebuffer framebuffer;
     VkDescriptorSet descriptor_set;
+} SubmissionResources;
+
+typedef struct {
+    VkImage image;
+    VkImageView view;
+    VkFramebuffer framebuffer;
+    VkSemaphore draw_complete_semaphore;
+    VkSemaphore image_ownership_semaphore;
 } SwapchainImageResources;
 
 struct demo {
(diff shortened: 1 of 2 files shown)


ziga-lunarg (approved):

water-chika:
> here are currently two submission resource sets, and the application alternates between the two on subsequent frames. The swapchain-image resources are separate because they are dependent on the number of swapchain images and because the swapchain image index used each frame isn't deterministic.

If there are only two submission resource sets, why we do not create a swapchain with two images?

water-chika:
> re-record the command buffers each frame instead of pre-recording them at application startup and during resize.

If two resource groups (per-submission, and per- swapchain-image) could be associated, the command buffers could still be pre-recorded.

Such as, if there are 4 swapchain-images and 2 submission-resource set:

| swapchain-image index | submission-resource index |
| -------------------------- | ------------------------------ |
| 0 | 0 |
| 1 | 1 |
| 2 | 0 |
| 3 | 1 |

And create 4 command buffers to record these.

charles-lunarg:
My stance is that pre-recording command buffers is not indicative of a typical program architecture, so makes for a poor demonstration of the Vulkan API because of how rare pre-recording can be used outside of simple demo applications (like vkcube). 

The other reason is that the order of swapchain images is 'effectively' (but not always) random, so to allow for continuous submissions you'd have to have a matrix of submission resource sets with each swapchain resource sets. And that makes understanding the synchronization flow significantly harder. The big idea is that we have a fence that protects the per-submission resources from being updated before the previous time the resource was used is done.

water-chika:
How about performance influence?

charles-lunarg:
No perceptible performance influence. Recording command buffers is CHEAP all things considered - and for complex applications the ability to change what is and isn't drawn offers significant performance advantages.

spencer-lunarg (changes requested):
I get a blank screen on both XCB and XLIB (didn't triage at all)

(But it is validation layer clean!)

charles-lunarg:
@spencer-lunarg good catch, I didn't run this on linux after making fixes on windows. It was a very simple inverted condition. 
```
if (!demo->initialized || !demo->swapchain_ready) { // this is backwards, it should be  if (demo->initialized && demo->swapchain_ready) 
            demo_draw(demo);
            if (demo->is_minimized) {
                demo->curFrame++;
            }
            if (demo->frameCount != INT32_MAX && demo->curFrame == demo->frameCount) demo->quit = true;
        }

spencer-lunarg:
`--wsi xlib` still give a blank white screen

charles-lunarg:
@spencer-lunarg :man_facepalming: it was the same issue xcb had, thanks for checking. I made sure to check if vkcubepp has the same problem, it does not.

spencer-lunarg (approved):
now works on xcb/xlib/wayland for both c and cpp

PR #1115, opened 2025-04-28
cube: link libdl for Linux

johnzupin:
Link DL lib. Volk used to do this but since it was removed an error will pop up on Ubuntu 20.04 for undefined references.


johnzupin:
Closing in favor of https://github.com/KhronosGroup/Vulkan-Tools/pull/1116

PR #1124, opened 2025-06-20
vulkaninfo: Prevent invalid json structures

charles-lunarg:
Prevents keys from being used when printing inside an array, as that isn't allowed by JSON.


charles-lunarg:
CI Failed with a known error - not going to hold up merging for it.

PR #1134, opened 2025-07-17
build: Remove pkg-config provided library names

charles-lunarg:
The library names of Xcb, Xlib, and Wayland do not need to be queried from pkg-config, instead they can be hardcoded to use the fallback names.

The intent of querying the library names was to prevent issues where the hardcoded name was not the platform-appropriate name. But because <library>_LINK_LIBRARIES can have more than one library name, the logic to assign <library>_LINK_LIBRARIES into a compile definition breaks horribly. While it is possible to handle this in CMake, the dlopen code would also have to handle it which is much more error prone.

### DIFF
--- a/cube/CMakeLists.txt
+++ b/cube/CMakeLists.txt
@@ -307,16 +307,6 @@ if (ANDROID)
     return()
 endif()
 
-if (XCB_LINK_LIBRARIES)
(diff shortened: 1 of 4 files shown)


charles-lunarg:
Going to take the lack of response as tacit approval - Can always fix things later if needed after all.

PR #1136, opened 2025-07-18
cube: Fix wayland resize broken when extent is 0xFFFFFFFF

charles-lunarg:
This is a 'magic value' that shouldn't be allowed to overwrite the current window width & height. There is code to handle the value, but not everywhere it can come from.


kanashimia (approved):
Fixes the problem for me on nvidia, don't see any difference on intel as it had no such problems. I didn't quite read the code with much thought, but should be good.

PR #1142, opened 2025-08-13
vulkaninfo: Default init VkVideoCapabilitiesKHR

charles-lunarg:
FIxes an issue seen running `vulkaninfo -j` on various windows systems in internal CI.


lunarpapillo:
Is this a fix, or a stopgap?  Initializing the data should stop the random output, but there's still an underlying problem that the data isn't being returned correctly, right?

charles-lunarg:
Yes, this isn't a full solution. I see that most of the variables used are initialized to zero, but it seems that VkVideoCapabilitiesKHR was missed. If the underlying issue is a driver not filling out the fields, then this PR helps isolate vulkaninfo from bad drivers. And even if it doesn't, zero-init doesn't hurt.

PR #1146, opened 2025-08-18
mock_icd: Remove legacy video support

aqnuep:
This is no longer needed now that Vulkan-Profiles supports video.


aqnuep:
> @aqnuep is this being removed because it is untested-and-wrong, or is there another reason
> 
> I guess if things are working, why remove it?

It's no longer maintained as it is subsumed by Vulkan-Profiles and currently it's already missing support for VP9 decode which was released _after_ Vulkan-Profiles got video support.

The entire code only existed to be able to test video related VVL code, but as now Vulkan-Profiles allow this to be entirely configured instead of being just a hard-coded set of caps, and VVL anyway does not use the Mock ICD for testing anymore (it has its own Test ICD, which similarly doesn't have hard-coded video caps, as there's the Vulkan-Profiles support).

PR #1153, opened 2025-09-09
 Use the clock_gettime() function instead of the obsolescent gettimeofday for macOS and iOS

Andreyogld3d:
Hi!
clock_gettime() can be used for macos since MacOS 10.12 Sierra and iOS 10 and never, instead of the obsolescent gettimeofday() function.

see POSIX documenation https://pubs.opengroup.org/onlinepubs/9699919799/functions/gettimeofday.html

### DIFF
--- a/cube/gettime.h
+++ b/cube/gettime.h
@@ -26,13 +26,9 @@
 
 #include <windows.h>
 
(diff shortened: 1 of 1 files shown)


aitor-lunarg (approved):
LGTM!

Issue #1138, opened 2025-08-03
cube: Wayland WSI is buggy


PancakeTAS:
After the Wayland WSI was made default in a recent commit, vkcube essentially stopped working unless I specified another backend.

Right after starting the process, a lot of validation errors pop up but no window. After a while a window appears and there's more warnings. Finally the window turns black and even more warnings appear.

I'm on Hyprland and I wasn't able to find any config option that affected this behavior, so default configuration should hopefully suffice for replication.

<details markdown="1">

<summary>Log</summary>

```
Selected WSI platform: wayland
Selected GPU 0: NVIDIA GeForce RTX 3080, type: DiscreteGpu
Validation Error: [ VUID-VkRenderPassBeginInfo-pNext-02852 ] | MessageID = 0xdb2c5767
vkCmdBeginRenderPass(): pRenderPassBegin->renderArea offset.x (0) + extent.width (1280) is greater than framebuffer width (500).
The Vulkan spec states: If the pNext chain does not contain VkDeviceGroupRenderPassBeginInfo or its deviceRenderAreaCount member is equal to 0, renderArea.offset.x + renderArea.extent.width must be less than or equal to VkFramebufferCreateInfo::width the framebuffer was created with (https://docs.vulkan.org/spec/latest/chapters/renderpass.html#VUID-VkRenderPassBeginInfo-pNext-02852)
Objects: 2
    [0] VkRenderPass 0xd000000000d
    [1] VkFramebuffer 0x270000000027

Validation Error: [ VUID-VkRenderPassBeginInfo-pNext-02853 ] | MessageID = 0x1bd44466
vkCmdBeginRenderPass(): pRenderPassBegin->renderArea offset.y (0) + extent.height (1399) is greater than framebuffer height (500).
The Vulkan spec states: If the pNext chain does not contain VkDeviceGroupRenderPassBeginInfo or its deviceRenderAreaCount member is equal to 0, renderArea.offset.y + renderArea.extent.height must be less than or equal to VkFramebufferCreateInfo::height the framebuffer was created with (https://docs.vulkan.org/spec/latest/chapters/renderpass.html#VUID-VkRenderPassBeginInfo-pNext-02853)
Objects: 2
    [0] VkRenderPass 0xd000000000d
    [1] VkFramebuffer 0x270000000027

Validation Error: [ VUID-VkRenderPassBeginInfo-pNext-02852 ] | MessageID = 0xdb2c5767
[...]

PR #1155, opened 2025-09-15
cube: Resize surface after toplevel configure

PancakeTAS:
It appears some compositors (tested on Hyprland) don't invalidate the swapchain after a window resize. This breaks rendering entirely due to mismatched sizes in the demo struct, versus the created framebuffer and isn't caught by anything else.


PancakeTAS:
This closes #1138, I should've put that into the description.

mahkoh:
The solution for this error from your logs is to double-buffer the width/height parameters sent in the xdg_toplevel.configure event until they are applied in the xdg_surface.configure event:

```
vkCmdBeginRenderPass(): pRenderPassBegin->renderArea offset.x (0) + extent.width (1280) is greater than framebuffer width (500).
```

charles-lunarg (approved):
Was able to run this and had no issue with resizing. Wayland code also makes more sense now, which is stellar.

PR #1159, opened 2025-09-19
build: Update to header 1.4.327

mikes-lunarg:


mikes-lunarg:
@charles-lunarg vulkaninfo video code was broken by the presence of a vendor extension. I have a workaround but I don't particularly like it. Any thoughts on the proper way to fix this?

PR #1160, opened 2025-09-20
cube: Fix xlib window close

ziga-lunarg:


charles-lunarg:
I don't mind the change, but could I get a 1 line description of what the problem is? Best to put it in the commit message.

charles-lunarg (approved):
Ahh this makes a lot of sense now! Thanks for the quick fix :D

PR #1163, opened 2025-10-01
cubepp: Fix uninitialized variables

ncesario-lunarg:


charles-lunarg:
I love uninitialized memory!

PR #1196, opened 2026-01-06
info: Add support for VK_KHR_display

jeremyg-lunarg:
This prints the results of vkGetPhysicalDeviceDisplayPropertiesKHR(), vkGetPhysicalDeviceDisplayPlanePropertiesKHR() and other related property lookups.

It does not provide support to create a VK_KHR_display surface and query it.

It also does not try to use VK_EXT_direct_mode_display to take control of the display from the window system.


charles-lunarg (approved):
PR looks reasonable. I did run it locally and got a reasonable output for zero supported displays, which is good.

PR #1227, opened 2026-06-05
Update to header v1.4.353

mikes-lunarg:


charles-lunarg (changes requested):
Uhhh seems like the vulkaninfo file was deleted by accident.

mikes-lunarg:
> Uhhh seems like the vulkaninfo file was deleted by accident.

Not really an accident, but codegen was crashing after opening the output file and the header bot was happy to commit an empty file.

scripts/generators/vulkaninfo_generator.py
```diff
@@ -712,14 +712,15 @@ def PrintGetFlagStrings(self,name, bitmask):
         out.append('    std::vector<const char *> strings;\n')
         # If a bitmask contains a field whose value is zero, we want to support printing the correct bitflag
         # Otherwise, use "None" for when there are not bits set in the bitmask
-        if bitmask.flags[0].value != 0:
```
mikes-lunarg, line 715: Needed a check here to prevent OOB list access for an empty bitmask type (VkFormatFeatureFlagBits4KHR in this case)

charles-lunarg (approved):

PR #1234, opened 2026-07-09
vulkaninfo: Make vulkan entry point variables static

hapakj:
Make vulkan entry point variables static in vulkaninfo.

Prevent link issues in case vulkan info directly linked to a library which exports the related vulkan entry points. e.g. a vulkan loader or and icd.


charles-lunarg (commented):
I don't understand the purpose of these changes.

Vulkaninfo is an application that *links* `vulkan-1.dll`/`libvulkan.so` and indirectly loads vulkan drivers. Why would the loader or a driver *link* to `vulkaninfo`? If you were doing that, there would need to be significantly more changes. 

Is this PR the result of a dynamic linker warning about duplicate symbols?

aqnuep:
> I don't understand the purpose of these changes.

> Vulkaninfo is an application that links vulkan-1.dll/libvulkan.so and indirectly loads vulkan drivers. Why would the loader or a driver link to vulkaninfo? If you were doing that, there would need to be significantly more changes.

> Is this PR the result of a dynamic linker warning about duplicate symbols?

@hapakj is on vacation. But the particular example is that some Vulkan SC drivers, for example statically link into the application, and the exported symbols conflict with the driver's symbols.

Although this can also happen with a dynamic linker due to the symbol conflicts.

charles-lunarg (approved):
Ah, I see the reason now. I should have suspected that what doesn't make sense for vulkaninfo does for vulkaninfoSC. 

As there isn't any drawback to adding static here, it is an easy approval.

PR #1239, opened 2026-07-31
Update to header v1.4.358

mikes-lunarg:


charles-lunarg (approved):
Oops, I thought I reviewed this already. My bad!

tgurr:
No sure if this is handled automatically and it just takes more time but the tag for v1.4.358 (https://github.com/KhronosGroup/Vulkan-Tools/releases/tag/v1.4.358) currently still appears to be missing as well.

charles-lunarg:
Thanks for catching this, it should be fixed now.

PR #1248, opened 2026-08-18
vulkaninfo: Include extended flags in video format props output

aqnuep:
@lolzballs, FYI.


lolzballs:
Thanks, this seems to work for the encode input/decode output/dpb formats but the quantization map formats don't have it still.

aqnuep:
> Thanks, this seems to work for the encode input/decode output/dpb formats but the quantization map formats don't have it still.

Hmm, interesting...

aqnuep:
@lolzballs, okay, should be fine now. Sorry about that.

lolzballs:
Yes, it works as expected now. Thanks!

charles-lunarg (approved):

